Issue #670, opened 2013-03-12
Rails 4 session values not persisting after redirect


beerlington:
I'm having [an issue ](http://stackoverflow.com/questions/15351985/session-not-persisting-across-redirect-with-omniauth-and-rails-4) where session values that I set in the SessionsController are not persisted when users are redirected to a different page.

In the following controller, I am setting the user_id and redirecting to a different controller's action. When execution gets to that controller, the value I previously set is no longer available. Similar logic works in a Rails 3.2 application so I'm not sure if it's a problem with OmniAuth or what.

``` ruby
class SessionsController < ApplicationController
  skip_before_filter :authenticate_user!

  def create
    user = User.find_or_create_by_auth_hash(auth_hash)
    session[:user_id] = user.id
    redirect_to root_path
  end

  protected

  def auth_hash
    request.env['omniauth.auth']
  end

end
```

jlkwarteng:
Currently Experiencing the very same issue. Although i'm not working with omniauth session variables i set do not persist to the next controller on redirect in rails

PR #674, opened 2013-03-27
Adding authenticity token so session will persist

beerlington:
This fixes issue #670
When using the developer login strategy sessions are not being persisted
in Rails 4.0.0.beta1 because the authenticity token is not present when
submitting the form. By adding the CSRF token to the form as a hidden
field, it can be verified, and sessions work properly.

### DIFF
--- a/lib/omniauth/form.rb
+++ b/lib/omniauth/form.rb
@@ -119,6 +119,10 @@ module OmniAuth
       self
     end
 
+    def authenticity_token_field(token)
+      @html << "\n<input type='hidden' name='authenticity_token' value='#{token}' />"
+    end
+
     def text_field(label, name)
       label_field(label, name)
       input_field('text', name)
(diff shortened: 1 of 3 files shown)


mltsy:
I have this same problem trying to implement a password grant flow omniauth strategy.  The request phase in this flow is just a form that gathers the password from the user and submits to the clalback.  When thinking about whether a CSRF token is really necessary in this form, I realized that the whole purpose of the form is to authenticate the user (they are entering their password on this form), so if some forged request has that information to submit from off the site, they've already totally compromised the user's security - a CSRF token isn't going to protect them.  Therefore I believe that skipping CSRF token verification is the right solution, at least for this particular case.

In the case of other grant flows (i.e. authorization code), the request should protected from request forgery using a [state parameter](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html), which is [specified in the OAuth spec](https://tools.ietf.org/html/rfc6749#section-10.12).  In this case, again, I believe Rails' built-in CSRF token verification would be redundant (if it even worked), and can therefore be skipped.

bradobro:
hey @mltsy would you please start a new issue and ref this one if it's pertinent.

mltsy:
Sorry, I was just verifying that this doesn't need to be fixed (with justifications of the workaround for various cases, for people who find this issue as I did) - I should have led with that :wink:  I'll update my comment!

PR #689, opened 2013-06-24
Generic error handling in Strategy

joshjordan:
Rescues from strategy calls by calling fail!, properly URL encodes message_key for FailureEndpoint handling

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -193,7 +193,7 @@ module OmniAuth
         return callback_call if on_callback_path?
         return other_phase if respond_to?(:other_phase)
       rescue => ex
-        return fail! ex.message, ex
+        return fail!(ex.message, ex)
       end
 
       @app.call(env)
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -186,10 +186,15 @@ module OmniAuth
       @env['omniauth.strategy'] = self if on_auth_path?
 
       return mock_call!(env) if OmniAuth.config.test_mode
-      return options_call if on_auth_path? && options_request?
-      return request_call if on_request_path? && OmniAuth.config.allowed_request_methods.include?(request.request_method.downcase.to_sym)
-      return callback_call if on_callback_path?
-      return other_phase if respond_to?(:other_phase)
+
+      begin
+        return options_call if on_auth_path? && options_request?
+        return request_call if on_request_path? && OmniAuth.config.allowed_request_methods.include?(request.request_method.downcase.to_sym)
+        return callback_call if on_callback_path?
+        return other_phase if respond_to?(:other_phase)
+      rescue => ex
+        return fail! ex.message, ex
+      end
 
       @app.call(env)
     end
--- a/spec/omniauth/strategy_spec.rb
+++ b/spec/omniauth/strategy_spec.rb
@@ -312,7 +312,9 @@ describe OmniAuth::Strategy do
       context 'disabled' do
         it 'does not set omniauth.origin' do
           @options = {:origin_param => false}
-          expect { strategy.call(make_env('/auth/test', 'QUERY_STRING' => 'return=/foo')) }.to raise_error('Request Phase')
+          strategy.should_receive(:fail!).with('Request Phase', kind_of(StandardError))
+
+          strategy.call(make_env('/auth/test', 'QUERY_STRING' => 'return=/foo'))
           expect(strategy.last_env['rack.session']['omniauth.origin']).to eq(nil)
         end
       end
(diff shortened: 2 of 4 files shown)


joshjordan:
@BobbyMcWho ?

BobbyMcWho:
@joshjordan this had been open for 8 years, 7 since last activity, so I closed it as stale. If you're still interested in this, feel free to rebase your changes onto current master and we can go from there

joshjordan:
@BobbyMcWho understood, but last action was on omniauth to review, which is why I was surprised to see it closed with no comment.

Updated to resolve conflicts with master.

BobbyMcWho:
This would be a breaking change, and I'm not confident that the value that it provides is worth adding the breaking change here. 

As it is, any failures in the strategy do not call `fail!` today, and an inherited strategy's `before_x_phase` and `x_phase` calls can already handle any errors that occur within them without bubbling them all the way up to base strategy code. 

Is there a specific use case you can provide an example of that handling the error within one of those inherited implementations isn't an option?

joshjordan:
Sure thing -- that's the encoded in the tests. The simplest example is a generic exception connecting to the third party service when trying to swap e.g. a token for auth credentials.

This isn't about strategies handling errors, its about the consumer being able to handle errors. We have the ability to set an `on_failure` endpoint, but notably, if there is an unhandled exception, the `on_failure` endpoint will not be called. OmniAuth will, in fact, just render the error top-level, meaning the user gets a 500 error. As a consumer, if the strategy has not handled the exception explicitly, it is extremely frustrating that there is currently *nothing* I can do to step my users from seeing errors, despite having properly configured and written a `FailureEndpoint` implementation according to the omniauth guidance.

> As it is, any failures in the strategy do not call fail! today,

And that's the problem, wouldn't you agree?. They just cause the entire endpoint to return a 500, which is unexpected behavior and not useful to anyone. I would be quite surprised if anyone desires that behavior, based on the dozens of omniauth developers I've spoken with over the years. Even if it were desired, it can still be achieved because this PR puts it in the consumer's control what ot do with the `fail!`. So maybe we'd call it a change in behavior, but perhaps not "breaking".

On that topic, I did a review of many of the most popular omniauth strategy implementations. Terribly few of them actually handle exceptions, and when I raised this issue with specific strategies' issue repos, the maintainers were surprised to find that omniauth was not already handling this. Critical implementations, such as `ominauth-oauth2` do handle the most common exception cases, but not all of them -- and the vast majority of omniauth strategies have 0 error handling in them. So, sure, they *could* do the error handling, but I think it is fairly safe to say at this point that in general, they simply do not. Should we really be asking 275+ maintainers to handle errors in 4 different methods, vs. catching exceptions and giving consumers a chance to have their application deal with them?

joshjordan:
To put it differently, this seems to me to be stricly an improvement. If you prefer, I could also allow gem consumers to opt-out of this functionality with a "no thanks, just blow up if there are exceptions" config flag.

lib/omniauth/strategy.rb
```diff
@@ -184,10 +184,15 @@ def call!(env) # rubocop:disable CyclomaticComplexity, PerceivedComplexity
...
-      return callback_call if on_callback_path?
-      return other_phase if respond_to?(:other_phase)
+
+      begin
+        return options_call if on_auth_path? && options_request?
+        return request_call if on_request_path? && OmniAuth.config.allowed_request_methods.include?(request.request_method.downcase.to_sym)
+        return callback_call if on_callback_path?
+        return other_phase if respond_to?(:other_phase)
+      rescue => ex
+        return fail! ex.message, ex
```
BobbyMcWho, line 194:
```suggestion
        return fail!(ex.message, ex)
```

spec/omniauth/strategy_spec.rb
```diff
@@ -312,32 +312,41 @@ def make_env(path = '/auth/test', props = {})
       context 'disabled' do
         it 'does not set omniauth.origin' do
           @options = {:origin_param => false}
-          expect { strategy.call(make_env('/auth/test', 'QUERY_STRING' => 'return=/foo')) }.to raise_error('Request Phase')
+          strategy.should_receive(:fail!).with('Request Phase', kind_of(StandardError))
```
BobbyMcWho, line 315: Can you change these to the `expect(strategy).to receive(:fail!)` syntax?

joshjordan:
I believe so — happy to make that change. Will do so after dinner.

BobbyMcWho:
You make good points, thanks for the detailed explanation.

Now is probably a good time to merge this since I will also be making breaking changes to close the CVE mentioned here: https://github.com/omniauth/omniauth/issues/960

I'm going to point these changes at the 2_0_indev branch.

BobbyMcWho:
Can you rebase onto 2_0_indev and resolve any conflicts?

joshjordan:
Sure thing! Should I repoint them merge target to that branch as well?

BobbyMcWho:
I went ahead and pointed them at it in the GitHub GUI

joshjordan:
@BobbyMcWho thanks! I updated the PR with the requested changes

BobbyMcWho (approved):

PR #735, opened 2014-04-10
Abstracting the session storage mechanism

ninjapanzer:
### What am I

We had a condition at ThinkThroughMath where an incoming request to the request phase included a rather large set of request params. This was causing a cookie overflow to occur. We do not want to incur the overhead of moving away from our cookie session_store. So I abstracted the session process in OmniAuth to accept a storage adapter. I have added the rack.session process back as the SessionStore adapter and create on for redis here https://github.com/ninjapanzer/omniauth-redis-store which hooks into the config process to allow me to define the redis client I want to use. After requiring the adapter config looks a little like this:

```
require 'omniauth-redis-store'

OmniAuth.config.request_store = OmniAuth::RedisStore
OmniAuth.config.redis = Redis::Namespace.new('OmniAuth', :redis => REDIS)
```
### The adapter spec

Adapters implement a Hash like interface with [], []=, and delete methods

I also moved the session method out of the strategy and into the concern of the storage adapter

### DIFF
--- a/lib/omniauth.rb
+++ b/lib/omniauth.rb
@@ -15,6 +15,7 @@ module OmniAuth
   autoload :Form,     'omniauth/form'
   autoload :AuthHash, 'omniauth/auth_hash'
   autoload :FailureEndpoint, 'omniauth/failure_endpoint'
+  autoload :SessionStore, 'omniauth/session_store'
 
   def self.strategies
     @strategies ||= []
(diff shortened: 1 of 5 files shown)


BobbyMcWho:
I'm going to close this as stale, it's a bit bigger of a change than I'm comfortable merging right now, but if this is still an issue then we can discuss it further via a new issue.

PR #742, opened 2014-05-04
Added chomp to on_path? for callback paths with trailing '/'

auser:
This PR fixes the cases where the `callback_path` requires a trailing slash, the `on_path?()` method.


tmilewski:
Apologies, I'm going to close this. Feel free to re-open if this is still problematic.

PR #755, opened 2014-06-25
log warn instead of error when user denied request

trungpham:
Log a warning instead of an error if the user denied the oauth request.


tmilewski:
Hey @trungpham,

1. Is this still an issue for you?
2. Keeping in mind that this is technically a breaking change, what's the full reasoning for this?
3. Please rebase on Master if you'd still like to consider this.

Thanks

PR #770, opened 2014-09-14
Developer strategy should use GET

maccman:


marr:
I am curious about this, too. In the docs, they say to add a `get` route https://github.com/intridea/omniauth#integrating-omniauth-into-your-application. 

If I add that get route, I obviously get a RoutingError:
```
INFO -- omniauth: (developer) Callback phase initiated.

ActionController::RoutingError (No route matches [POST] "/auth/developer/callback"):
```

BobbyMcWho:
We will no longer recommend GET routes as of 2.0, so I am closing this.

PR #772, opened 2014-09-25
Fixing issues with turbolinks

silasb:
Turbolinks will not load the CSS in the head of the document.

Moving to the body will make Turbolinks work again.


tmilewski:
Is this still an issue? If so, would you mind rebasing on Master?

silasb:
I don't know if it's still an issue or not. I haven't touched that code in
two years.  We can close this ticket if you want.

tmilewski:
Yeah, I'm going to close for now. Should it pop back up, we'll address it then. Thanks a lot!

PR #779, opened 2014-10-28
Pass through test_mode with another request method

inkstak:
Hello,
In one of my rails app, I use Omniauth with multiple providers.
For each one, I can disconnect by calling a DELETE on `auth/:provider`

```
GET|POST /auth/:provider(.:format)            identities#passthru
GET|POST /auth/:provider/callback(.:format)   identities#callback 
DELETE   /auth/:provider(.:format)            identities#destroy
```

It works fine except on `test_mode`, where destroy requests are mapped as request phases.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -257,7 +257,7 @@ module OmniAuth
     # in the event that OmniAuth has been configured to be
     # in test mode.
     def mock_call!(*)
-      return mock_request_call if on_request_path?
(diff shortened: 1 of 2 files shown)


r7kamura:
+1 for checking HTTP method :+1:
I'm troubled in the same problem in my app that has `DELETE /auth/:provider` endpoint too. I couldn't add tests for this endpoint because omniauth took over the request and return mock response :weary:

takashi:
any updates on this PR? @sferik

PR #806, opened 2015-05-13
Sets omniauth.headers on the request phase

### LINKED ISSUE #804: Add something like `env["omniauth.headers"]`
In the same way that params are kept in `env['omniauth.params']`, it would be desirable to keep headers in `env['omniauth.headers']`.

I'm trying to implement multiple providers. The problem is that I'm not using a cookie based authentication system, but a token one, where users are sending their credentials in the request headers each time.

The problem is that I loose that credentials when the user goes back to my app from the provider.

For example, say a user calls `/auth/github` with its credentials in the headers. Then, from `/auth/:provider/callback` I can't retrieve them so I can't say if the user was logged in or not.

It would be great if omniauth had some way to keep headers info, and adding `env['omniauth.headers']' seems the most natural for me .

Thanks

waiting-for-dev:
Add `env['omniauth.headers']` which is a hash with header values on the request phase. Closes #804

### DIFF
--- a/spec/omniauth/strategy_spec.rb
+++ b/spec/omniauth/strategy_spec.rb
@@ -706,7 +706,7 @@ describe OmniAuth::Strategy do
       it 'sets omniauth.headers on the request phase' do
         OmniAuth.config.mock_auth[:test] = {}
 
-        strategy.call(make_env('/auth/test', { 'HTTP_FOO' => 'bar' }))
+        strategy.call(make_env('/auth/test', 'HTTP_FOO' => 'bar'))
         expect(strategy.env['rack.session']['omniauth.headers']).to eq('FOO' => 'bar')
       end
 
--- a/spec/omniauth/strategy_spec.rb
+++ b/spec/omniauth/strategy_spec.rb
@@ -703,6 +703,13 @@ describe OmniAuth::Strategy do
         expect(strategy.env['rack.session']['omniauth.params']).to eq({})
       end
 
+      it 'sets omniauth.headers on the request phase' do
+        OmniAuth.config.mock_auth[:test] = {}
+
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
Hey, new to Travis, but it seems all tests are ok but there is some code offenses not introduced by myself.

zhen639:
@waiting-for-dev How's your PR going?  Your PR is pretty helpful for many people.

waiting-for-dev:
Hey, THIS is the pull request and it has been ready for a long time waiting for some feedback ;)

spec/omniauth/strategy_spec.rb
```diff
@@ -654,6 +654,13 @@ def make_env(path = '/auth/test', props = {})
         expect(strategy.env['rack.session']['omniauth.params']).to eq('foo' => 'bar')
       end
 
+      it 'sets omniauth.headers on the request phase' do
+        OmniAuth.config.mock_auth[:test] = {}
+
+        strategy.call(make_env('/auth/test', 'HTTP_FOO' => 'bar'))
```
tmilewski: Mind testing that it doesn't pick up env's that aren't prepended with 'HTTP_'?

tmilewski:
@waiting-for-dev Would you mind rebasing on Master?

spec/omniauth/strategy_spec.rb
waiting-for-dev: Well, this PR was just intended to keep header values, which in rack env are always prepended with `HTTP_`.

waiting-for-dev:
Hey @tmilewski . Rebase done. Please, review it is correct, it has been a long time since I visited it.

PR #809, opened 2015-05-25
Protect request phase against CSRF when Rails is used.

DouweM:
TL;DR, as of 2020-03-07: 

> Please note: there is currently a CSRF vulnerability which affects OmniAuth (designated [CVE-2015-9284](https://nvd.nist.gov/vuln/detail/CVE-2015-9284)) that requires mitigation at the application level. More details on how to do this can be found on the [Wiki](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284).

Original issue description:

---

The request phase in OmniAuth is currently vulnerable to Cross-Site Request Forgery, which allows an attacker to easily gain full access to a user's account on a site that uses OmniAuth, when used in combination with another CSRF vulnerability on the side of a connected OAuth provider.

This vulnerability was reported to us (GitLab) by Mohamed Abdelbaset Elnoby (@symbiansymoh), a Senior Information Security Analyst at [Seekurity.com](http://seekurity.com), on April 23rd. On April 24th, we reported it to the OmniAuth team via email. On May 14th, we [merged our patch](https://gitlab.com/gitlab-org/gitlab-ce/commit/c2ee828c19cb245809647428334b8ef215536a0d) into the public GitLab repository.

Mr. Elnoby also gave us an example of an OAuth provider that had the CSRF vulnerability on their side that is needed to make this attack "weaponizable". The vulnerability has been reported to them.

The below description of the issue is written from GitLab’s perspective.

> Connecting the GitLab account to an OAuth provider like Google works like this:
> 1. User visits https://gitlab.com/users/auth/google
> 2. User is redirected to a page on Google
> 3. Google checks if the user is logged in
> 4. If the user is logged in, Google checks if the user has authorized the GitLab.com application to use their Google account
> 5. If the user has authorized the GitLab.com application to use their Google account, Google redirects back to GitLab
> 6. GitLab adds the Google account to the GitLab account as an alternate identity, which means the Google account can be used to log into the GitLab account.
> 
> This is all very straightforward, and works as expected.
> 
> Note that if the Google user has already authorized the GitLab.com application to use their Google account, there is no user action needed to link the two accounts, apart from the initial https://gitlab.com/users/auth/google visit.
> 
> This means that if you can get a person who 1) is signed into Google, 2) has authorized the GitLab.com application to use it, and 3) is signed into GitLab.com, to visit https://gitlab.com/users/auth/google, their account will be linked to the Google account without any further interaction, and without any feedback to the user.
> 
> This is not a big deal if the signed in Google account belongs to the GitLab user themselves, but it is if the user is somehow signed into Google using an account belonging to an attacker. In this scenario, the GitLab account would be linked to the attacker's Google account, and the GitLab user would be none the wiser.
> 
> And let it just so be the case that that iss possible with _Redacted_: there exists a vulnerability on their side that allows an attacker to log into _Redacted_ using CSRF, which means that if a user visits a specially crafted page, they will end up being logged into _Redacted_ as the attacker.
> 
> Combining this with the described GitLab.com flow, forcing the victim to visit that specially crafted page, and directly after that, forcing them to visit https://gitlab.com/users/auth/redacted (by using an iframe, for example), will give the attacker full control over the victim's GitLab account. "Forcing them" can be as simple as sending them a link over email and getting them to click it.

To actually use the POST method rather than GET so the CSRF token is sent along, we have changed our links to the auth path from:

``` ruby
link_to "Log in with #{provider}", omniauth_authorize_path(:user, provider)
```

to:

``` ruby
link_to "Log in with #{provider}", omniauth_authorize_path(:user, provider), method: :post
```

Note that the current implementation is very specific to Rails. It needs to be, as far as I can see, since the CSRF protection needs to use the web framework’s authenticity token. 

I suggest deprecating `OmniAuth.config.allowed_request_methods` and only allowing `:post` from now on, but I haven't made that change yet since I think it warrants more discussion.

People that use OmniAuth with Rails will need to change the `link_to` helpers and `<form>` tags in their own code to use `POST` instead of `GET`, 
so I’m afraid this is a breaking change and can’t just go in a patch release, in SemVer terms.

cc @sferik

### DIFF
--- a/lib/omniauth.rb
+++ b/lib/omniauth.rb
@@ -2,6 +2,8 @@ require 'rack'
 require 'singleton'
 require 'logger'
 
+require 'omniauth/railtie' if defined?(Rails)
+
 module OmniAuth
   class Error < StandardError; end
 
--- a/lib/omniauth/railtie.rb
+++ b/lib/omniauth/railtie.rb
@@ -1,0 +1,15 @@
+require 'omniauth/request_forgery_protection'
+
+module OmniAuth
+  class Railtie < Rails::Railtie
+    initializer 'omniauth' do
+      OmniAuth.config.logger = Rails.logger
+
+      # Protect request phase against CSRF.
+      OmniAuth.config.allowed_request_methods = [:post]
+      OmniAuth.config.before_request_phase do |env|
+        RequestForgeryProtection.new(env).call
+      end
+    end
+  end
+end
--- a/lib/omniauth/request_forgery_protection.rb
+++ b/lib/omniauth/request_forgery_protection.rb
@@ -1,0 +1,67 @@
+require 'action_dispatch/http/request'
+
+module OmniAuth
+  # Based on ActionController::RequestForgeryProtection.
+
+  class RequestForgeryProtection
+    def initialize(env)
+      @env = env
+    end
+
+    def request
+      @_request ||= ActionDispatch::Request.new(@env)
+    end
+
+    def session
+      request.session
+    end
+
+    def reset_session
+      request.reset_session
+    end
+
+    def params
+      @_params ||= request.parameters
+    end
+
+    def call
+      verify_authenticity_token
+    end
+
+    def verify_authenticity_token
+      return if verified_request?
+
+      OmniAuth.logger.warn "Can't verify CSRF token authenticity" if OmniAuth.logger
+      handle_unverified_request
+    end
+
+  private
+
+    def protect_against_forgery?
+      ::ApplicationController.allow_forgery_protection
+    end
+
+    def request_forgery_protection_token
+      ::ApplicationController.request_forgery_protection_token
+    end
+
+    def forgery_protection_strategy
+      ::ApplicationController.forgery_protection_strategy
+    end
+
+    def verified_request?
+      !protect_against_forgery? || request.get? || request.head? ||
+        form_authenticity_token == params[request_forgery_protection_token] ||
+        form_authenticity_token == request.headers['X-CSRF-Token']
+    end
+
+    def handle_unverified_request
+      forgery_protection_strategy.new(self).handle_unverified_request
+    end
+
+    # Sets the token value for the current session.
+    def form_authenticity_token
+      session[:_csrf_token] ||= SecureRandom.base64(32)
+    end
+  end
+end


DouweM:
The Travis CI build fails because of RuboCop offenses, most of them not caused by this PR. Should I fix all of them?

sferik:
@DouweM I’m fixing them on `master` now. You should be able to rebase after that to make the tests pass.

DouweM:
Thanks Erik.

sferik:
Okay, the build on master should be passing. Try rebasing.

DouweM:
@sferik Build is green!

sferik:
Thanks for this patch. Normally, I’d merge and release a security patch immediately but I’m reluctant to do so in this case for two reasons:

1. this is a breaking change and
2. it contains Rails-specific code

I’d propose that we solve this by creating a new gem: `omniauth-rails`, that depends on `omniauth` and `rails` and contains all Rails-specific code. We could release this as version 1.0.0 and notify users of OmniAuth that they should depend on `omniauth-rails` instead of `omniauth` if they’re using OmniAuth with Rails.

I’d like to get :+1: from the original author of OmniAuth (@mbleigh) before making this change as well as the Rails security team (@tenderlove, @jeremy, @NZKoz).

If everyone agrees this is a good idea, we can move forward with this plan very quickly.

DouweM:
@sferik Moving to `omniauth-rails` makes sense. We'd need to get devise to move on this as well.

sferik:
@DouweM That should be doable. Ping: @josevalim and @carlosantoniodasilva.

sferik:
Two more questions:

1. Has a CVE already been assigned for this vulnerability?
2. When I click on Seekurity.com, I get a GoDaddy parked page. Is that really the security firm with which @symbiansymoh is associated?

DouweM:
1. No, I haven't done anything on that. From our earlier interaction, I wasn't sure if the issue was as serious as I thought and if it warranted that.
2. He says so.

sferik:
> 1. No, I haven't done anything on that. From our earlier interaction, I wasn't sure if the issue was as serious as I thought and if it warranted that.

@DouweM I think it would be helpful to have a CVE-ID for this vulnerability, so we can reference it unambiguously. Since you are most familiar with this issue, could you please initiate that process? The process to request a CVE-ID is described [here](https://cve.mitre.org/cve/request_id.html).

DouweM:
@sferik Thanks, I'll go through the motions.

sferik:
> @sferik Thanks, I'll go through the motions.

@DouweM Thank you.

I’ve contacted the Rails Security Team, informing them of this issue and asking for their help in spreading the word once it is patched.

DouweM:
A CVE request has been submitted to the [oss-security mailing list](http://www.openwall.com/lists/oss-security/). I'll link to it here when it becomes available on the site.

tenderlove:
I guess I don't understand how this is a Rails specific issue.  Any framework you use with Omniauth will have this problem if you're using GET requests (though this solution is for Rails apps).  This patch seems fine for handling the Rails side (though I thought we'd do everything automatically if you did a `link_to` with a `method: post`)

rafaelfranca:
@tenderlove I think this issue is not Rails specific so we don't need to handle it, @sferik only want our review in his plan.

The plan looks good. The Rails team can help to spread the word via our twitter account and the security list.

tenderlove:
> @tenderlove I think this issue is not Rails specific so we don't need to handle it, @sferik only want our review in his plan.

Gothcha.  It seems good to me.  The only question I have is since the link is turned in to a form, wouldn't we already take care of CSRF validation?  I don't know omniauth internals.  I assume it's a middleware?  If Omiauth was placed after we do CSRF checking, couldn't we delete most of the code in this patch?

DouweM:
> Gothcha. It seems good to me. The only question I have is since the link is turned in to a form, wouldn't we already take care of CSRF validation? I don't know omniauth internals. I assume it's a middleware? If Omiauth was placed after we do CSRF checking, couldn't we delete most of the code in this patch?

OmniAuth intercepts the request at the Rack middleware level, before it gets to the Rails controller and the `protect_against_forgery` before_action at all. If Rails had a `RequestForgeryProtection` middleware, that would be useful here, but that wouldn't work for regular Rails controllers, since it depends on controller-level settings.

tenderlove:
> OmniAuth intercepts the request at the Rack middleware level, before it gets to the Rails controller and the protect_against_forgery before_action at all. If Rails had a RequestForgeryProtection middleware, that would be useful here, but that wouldn't work for regular Rails controllers, since it depends on controller-level settings.

Makes sense.  :+1: from me on this patch. :)

DouweM:
@sferik So what's the next step? If you create an `intridea/omniauth-rails` repo, I'm happy to build a gem out of this patch and move it there. I think it's better to have it under `intridea` than `DouweM` or `gitlabhq`, so people will know it's the "official" OmniAuth+Rails gem.

sferik:
@tenderlove @rafaelfranca Thanks for taking the time to review this patch.

@DouweM I just created an empty repo at https://github.com/intridea/omniauth-rails. I’ll initialize that with a gem scaffold now. Feel free to open a pull request against that repo with the same code you’ve placed here. I’ll be busy for the next few hours but will try to push out a release later on this evening.

The `README` of this repo should also be updated to indicate that Rails users should depend on `omniauth-rails` instead of directly on `omniauth`.

I’m going to close this pull request.

DouweM:
@sferik All right. I won't have time tonight, unfortunately, but I'll submit a pull request tomorrow.

SymbianSyMoh:
Hi Guys,
Sorry for being late involving this thread, so many things happening here.
@sferik  Yup, that's exactly the domain but not yet stared and the website is fully under constructions.

DouweM:
Hey @SymbianSyMoh, thanks for finding this vulnerability and reporting it to us! 

I hope you don't mind that that I've taken up the task of reporting where relevant and making sure it gets patched. I've credited you here and in the CVE request: http://www.openwall.com/lists/oss-security/2015/05/26/11.

SymbianSyMoh:
Thanks @DouweM really appreciate that :+1:

SymbianSyMoh:
@DouweM @sferik @tenderlove Should I report this issue under this responsible disclosure rules:
1. https://hackerone.com/ruby
2. https://hackerone.com/rails
Or not ?!

rafaelfranca:
@SymbianSyMoh for this case no, it is not related to Ruby or Rails projects. But if you find something on one of these projects, those are the recommended channels.

SymbianSyMoh:
Thanks @rafaelfranca Appreciate it

DouweM:
Pull request agains omniauth-rails submitted: https://github.com/intridea/omniauth-rails/pull/1

SymbianSyMoh:
Hi Guys, I know this is a long time ago thread, but guess what? I never
received any CVE regarding this issue till now!! So what can i do to get it?



*Mohamed Abdelbaset Elnoby*Guru Programmer, Senior Information Security
Consultant & Web Application Penetration Tester at Seekurity Inc
<http://www.Seekurity.com>.

Contact me at:
LinkedIn
<https://www.linkedin.com/in/symbiansymoh>Facebook
<https://fb.com/symbiansymoh>Twitter <https://twitter.com/symbiansymoh>
<https://twitter.com/symbiansymoh>

DouweM:
@SymbianSyMoh I never got a response to the CVE request on the oss-security mailinglist, unfortunately.

reedloden:
I requested a CVE from the DWF (cc @kurtseifried) for this.

SymbianSyMoh:
So what is the problem?



*Mohamed Abdelbaset Elnoby*Senior Information Security Analyst & Chief
Product Security at Linio LATAM <https://www.linio.com.mx>, Founder at
BugBountyProgram
Platform <http://www.BugBountyProgram.com>, Security Advisor & Web
Application Penetration Tester at Seekurity <http://www.Seekurity.com>.

Contact me at:
LinkedIn
<https://www.linkedin.com/in/symbiansymoh>Facebook
<https://fb.com/symbiansymoh>Twitter <https://twitter.com/symbiansymoh>

*Let's speak encrypted. Here's my Public Key on PGP Global Directory: **Click
to download*
<https://keyserver2.pgp.com/vkd/DownloadKey.event?keyid=0x7F70C8FCAE73E1ED>

sdogruyol:
Any update on this? 

Fast forward to 2019, this is still valid. Any Ruby on Rails application using default `Omniauth`settings are open to this vulnerability. 

@reedloden @SymbianSyMoh since @DouweM had already opened a CVE, how do we proceed now?

SymbianSyMoh:
@sdogruyol +1, We haven't received any CVE updates till the moment!

reedloden:
I don't think a CVE was ever assigned, correct? I've assigned this CVE-2015-9284.

https://github.com/omniauth/omniauth-rails/pull/1 still needs to be merged and a new version of `omniauth-rails` released.

I don't feel comfortable adding this to ruby-advisory-db until we have a fixed version, especially considering how old this is. Would just make a lot of people very grumpy.

Can somebody take ownership of getting the actual fix into a released version?

sdogruyol:
@reedloden is there a link to assigned CVE-2015-9284?

reedloden:
It’ll be at
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-9284 once MITRE
updates the entry. Update request was sent in last week. I know they are
pretty backlogged right now though.

However, I can confirm that that CVE ID is correct and active. You can go
ahead and start using it.

will:
This cve has made it into github's dependency scanner thing.

Reading through the comments here I'm confused if this is rails specific, or any apps using omniauth. It seems like it's all apps, but the new CVE singles out rails.

Is there a summary of what we need to do for both rails and also non-rails apps to resolve this issue? It seems like there are no patched versions to upgrade. Is the resolution instead configuration changes?

zachfeldman:
GitHub just opened a security alert for this 3 year old vuln on our repository, with no patch version to upgrade to indicated. Just checking in as I'm a little unclear from the conversation if there's an obvious patch for this, or if there's still one being settled on. I'm sure others will come here soon with similar questions. Thanks for all the work thus far!

tmilewski:
It's been a long while since I've touched the core code here or even actively used Omniauth. Bear with me.

It seems to me that, for Rails, the approach (as defined by Gitlab) is to do the following:

1. Add `method: :post` to any links utilizing `link_to omniauth_authorize_path ...`:
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/views/devise/shared/_omniauth_box.html.haml#L8
2. Force `allowed_request_methods` to only allow `:post`:
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config/initializers/omniauth.rb#L13
3. Implement `RequestForgeryProtection`
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/gitlab/request_forgery_protection.rb
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/spec/lib/gitlab/request_forgery_protection_spec.rb
4. Add a `before_request_phase` which calls `RequestForgeryProtection`
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config/initializers/omniauth.rb#L16-18

This _obviously_ isn't a solution. 

What I'd like to understand at the moment is how far-reaching this CVE is?

Per @will, if it is further than just Rails, I'm unsure of the best way to approach the issue, especially without breaking every current implementation.

sikachu:
We're doing similar thing to @tmilewski as well here at Cookpad. We had to change all of the link to `/auth/[provider]` to go through HTTP POST (`button_to` helper works like a charm) and have this verifier in our lib directory:

**lib/request_forgery_protection_token_verifier.rb**
```ruby
# Provides a callable method that verifies Cross-Site Request Forgery protection
# token. This class includes `ActionController::RequestForgeryProtection`
# directly and utilizes `verified_request?` method to match the way Rails
# performs token verification in Rails controllers.
#
# If you like to learn more about how Rails generate and verify authenticity
# token, you can find the source code at
# https://github.com/rails/rails/blob/v5.2.2/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L217-L240.
class RequestForgeryProtectionTokenVerifier
  include ActiveSupport::Configurable
  include ActionController::RequestForgeryProtection

  # `ActionController::RequestForgeryProtection` contains a few configurable
  # options. As we want to make sure that our configuration is the same as what
  # being set in `ActionController::Base`, we should make all out configuration
  # methods to delegate to `ActionController::Base`.
  config.each_key do |configuration_name|
    define_method configuration_name do
      ActionController::Base.config[configuration_name]
    end
  end

  def call(env)
    @request = ActionDispatch::Request.new(env)

    unless verified_request?
      raise ActionController::InvalidAuthenticityToken
    end
  end

  private

    attr_reader :request
    delegate :params, :session, to: :request
end
```

And in our initializer file we have:

**config/initializers/omniauth.rb**
```ruby
require "request_forgery_protection_token_verifier"

  # ...
  OmniAuth.config.allowed_request_methods = [:post]
  OmniAuth.config.before_request_phase = RequestForgeryProtectionTokenVerifier.new
end
```

So, similar to Gitlab, but I tapped directly into `ActionController::RequestForgeryProtection` module instead of creating a controller.

This is against Rails 5.2. I have a feeling we could extract this as a gem so we can easily share our mitigation code and have test coverage against multiple versions of Rails.

tmilewski:
Using the information provided above, I have a quick proof-of-concept. While the specs pass, I haven't been able to test it in an actual Rails application. Feel free to leave any feedback over there.

https://github.com/omniauth/omniauth-rails/pull/2

tmilewski:
My thought process, let me know what you think...

1. **Merge https://github.com/omniauth/omniauth-rails/pull/2 to `omniauth-rails` master**
This gem will force allowed methods to `:post` and validate the request phase. As it stands, it will look for the `validate_request_phase` hook in OmniAuth (not yet implemented) and run `RequestForgeryProtection` there. If that doesn't exist, it will run in the `before_request_phase` hook. The intent here is that it (versions >1.9 of OmniAuth) will be backward-compatible with developer's existing `before_request_phase` hooks.
2. **Add `validate_request_phase` hook to OmniAuth**
3. **Add warnings**
Check if a user is using Rails without the OmniAuth Rails gem and *warn* them that they are subject to this CVE. *<-- I'm very open to suggestions on how to handle this approach.*

**Side Discussion:**
`omniauth-rails` gem already exists. Should I publish as `omniauth-ror`?

sikachu:
@tmilewski I'm sorry, I just noticed that you already worked on a PR for the mitigation. I was spending my work day extracting code from our project, writing test, and release a gem ... and I forgot to check back here.

So, I extracted our code and released [omniauth-rails_csrf_protection](https://rubygems.org/gems/omniauth-rails_csrf_protection) gem. The source code along with tests are at https://github.com/cookpad/omniauth-rails_csrf_protection.

I'm running the test on [CircleCI](https://circleci.com/gh/cookpad/omniauth-rails_csrf_protection/tree/master) against Ruby 2.5, 2.6 and Rails 5.1, 5.2, 6.0 RC, and edge Rails. (The workflow part is not visible, but you should be able to get to the separate builds using the link above.)

Would this be an acceptable solution for this mitigation? I'll be happy to update the code to use `validate_request_phase` when the time comes as well.

tmilewski:
@sikachu I appreciate the work! Thank you! If you don't mind testing on `Ruby 2.4` as well, considering that's it's not EOL, that'd be great.

I'm going to work on getting the `validate_request_phase` implemented as well as the warning (if you have any thoughts on that, please speak up). If you can, just implement something similar to this: https://github.com/omniauth/omniauth-rails/blob/28473b05d76a61b58da8a86e296e2ee728af4245/lib/omniauth-rails/railtie.rb#L10-L16 so that it'll work on versions <=1.9.

Considering the nature of the work and its impact, I'd like to move this repo over to the `omniauth` org and I can make you an owner of that.

jrochkind:
Hi! Why is there an extra gem you need to install to mitigate the CVE, instead of a fix being released in omniauth? This seems confusing to me, and makes me wonder about the maintenance story of omniauth...

Or is this just a temporary thing while you get it all worked out in omniauth (hopefully quickly?)

If not just a temporary thing... An extra gem seems like added maintenance burden to keep it in sync as omniauth potentially changes (which results in potential added dependency management burden down the road for developer users, to make sure they have the csrf_protection gem which works with the right version of omniauth), and is also not obvious to someone installing omniauth that they *also* need to install another gem if they want *secure* omniauth. This seems odd.

CHTJonas:
I whole-heartedly agree with what @jrochkind said. I think having the two gems is just confusing and there will be people who inadvertently use OmniAuth without the CSRF protection bit and get caught out.

I'd argue that the CVE is still valid and unfixed until a PR gets merged into this repo that addresses the issue generically for any given Rack application - not just people who use Rails. I'm sure there's a non-zero number of Sinatra users out there who would appreciate a fix! If it's a case of "we need to get this sorted ASAP even if it's just for Rails apps for the time being" then I'd propose just making the Rails-specific changes here and doing a 'temporary' release before fixing the issue generally and doing a proper release. The `omniauth-rails_csrf_protection` gem seemingly just adds some Railties that could be ignored by non-Rails apps no?

j1mmie:
> Why is there an extra gem you need to install to mitigate the CVE, instead of a fix being released in omniauth?

@jrochkind let the dust settle a bit. Takes a while to understand the implications of security changes (and breaking changes). @sikachu was kind enough to quickly offer a fix. Take it or leave it.

lolk1:
> > Why is there an extra gem you need to install to mitigate the CVE, instead of a fix being released in omniauth?
> 
> @jrochkind let the dust settle a bit. Takes a while to understand the implications of security changes (and breaking changes). @sikachu was kind enough to quickly offer a fix. Take it or leave it.

This issue has been open for over 4 years. The dust has not only settled, it probably has about 5 layers by now. Their concerns are valid and the passive aggressive "take it or leave it" statement is not helpful towards a proper resolution.

tmilewski:
Let's all take a breather. I wasn't on the project at the time this surfaced, nor did I triage the ticket, and I'm working on picking up the pieces on something I haven't used day-to-day for quite a while.

---

You're all right; however, there are a few issues here, and it's not as straight forward as it may seem.

OmniAuth is in use with a vast number of frameworks, gems, and provider. As such, us shoehorning each of these possibilities into OmniAuth core is entirely untenable in the short and long-term. It's the exact reason provider gems exist.

Additionally, per the CVE, the CSRF update needed requires front-end changes for passing tokens via forms. In the case of Rails, this is handled simply by `link_to ..., method: :post` but it's not always going to be that simple. It feels like maintaining framework-specific front-end code is a losing proposition.

The approach we seem to have right now is to do the following:

- OmniAuth 2.0.0: Implement the ability to tie validation into OmniAuth without breaking `before_request_filter` functionality
- Create OmniAuth gems which service the needs of each framework and situation (not necessarily just CSRF). 
- OmniAuth 2.0.0: Default `allowed_request_methods` to `[:post]` and warn on a lack of validation with suggestions to mitigate the issue.

The new specific gems would supersede the `omniauth` core gem in the same way the provider gems do.

---

(All of this said, I'm open to suggestions and, even better, PRs.)

CHTJonas:
Hey it's open source - really appreciate what you and @sikachu have done in such a short space of time! ❤️ I see what you're getting at, and to be honest maybe I underestimated the complexity of the changes each separate framework will need. Your roadmap above sounds sensible, just a shame someone's already taken `omniauth-rails` as a gem name.

tommyalvarez:
@tmilewski  I see you're working on an open PR on omniauth-rails (https://github.com/omniauth/omniauth-rails/pull/2). Once finished, will that be the official 'patch' to address this CVE issue? So I subscribe to the PR and get notified once it's merge and the Gem officially released. Let me know and thank you very much for your work!

lacostenycoder:
Do we know yet if devise is onboard with this and/or ominauth-rails  yet?  Thanks everyone

nbibler:
> Do we know yet if devise is onboard with this and/or ominauth-rails yet? Thanks everyone

Good question. I would guess that many (most?) Rails applications which use OmniAuth do so through a Devise integration. It would be useful to loop @plataformatec in. Whether the fix gets implemented directly into Devise (maybe this is preferable?) or the Devise OmniAuth integration gets included through omniauth-rails, I don't know that it would make much difference. Either way, some combination of the code, the documentation, and the examples in Devise will need to be updated to alert and account for the necessary changes.

sikachu:
@tmilewski I've updated the gem to support Ruby 2.4, and @GUI was kind enough to submit a PR to make sure we build against Rails (actionpack) 4.2 since it's still being supported for security patch.

I need to check with the right people here if it's okay for me to move repo ownership to OmniAuth since it's consider company's property. While the code is okay to be open source, the code is still under our organization so we "own" the code somewhat. I'll get back to you on that asap.

And yeah, I understand that it's not optimal to having to install another gem to enable mitigation in OmniAuth. However, the mitigation code definitely do not belong in OmniAuth gem as OmniAuth is actually framework agnostic. The only way we could ship this with OmniAuth would be that if someone extract the strong_parameters key generation and verification part into a Rack-based gem, then maybe we can include that. At this point, I think what we have right now is adequate enough.

a-ta-ta:
I'm using omniauth only as part of devise. Am I supposed to just wait when devise updates?

blindMoe:
> I'm using omniauth only as part of devise. Am I supposed to just wait when devise updates?
> Maybe it makes sense to start using omniauth-rails immediately if it doesn't conflict with devise?
> Have anybody tried to do so?

It doesn't look like omniauth-rails is ready yet

glenpike:
So we have implmented the omniauth-rails_csrf_protection solution, but previously we had our 3rd party OAuth provider log people in after they had verified the registration.  This would now require them POSTing to our `/auth/provider` endpoint with a CSRF token, which is not possible as they are on a separate platform system.
I'm guessing that this type of flow is one that could potentially be affected by the exploit?
Apart from implementing a 'landing page' on our site for newly registered users that uses JavaScript to POST to the `/auth/provider` - does anyone have any other suggestions to 'fix' the flow?

SymbianSyMoh:
Wow huge efforts been made here thanks for everyone. 

Funny thing is that i have reported this issue responsibly which affects a wide-range of 3rd party libraries and project without getting a dime :/

@DouweM  I'm not sure if such issue is eligible for a bounty either from @gitlab or any other organization else. I know asking for a bounty is something lame or at least i believe so!

Cheers!

CHTJonas:
@SymbianSyMoh looks like GitLab do have a bug bounty program - https://hackerone.com/gitlab might be worth a short retrospectively.

abrom:
For anyone looking for a more generic Rack app solution (in my case Sinatra), I've detailed mine here:

https://gist.github.com/abrom/effe58b27a4f4ac1b97fb85593d7c3be

Uses the Sinatra `rack-protection` gem for the CSRF checks as a base, as well as the processes from the Rails solutions outlined above.

iainbeeston:
Should the omniauth readme be updated to mention that anyone using omniauth with rails should also use `omniauth-rails_csrf_protection`?

eshnil:
Having to `POST` to `/auth/auth0` would pose a problem in cases where the user wants to access a protected page, and the app redirects the user (which can only be followed by a `GET`) to login page. It seems all those redirects would break with a 404 `No route matches [GET] "/auth/auth0"` error.

glenpike:
> Having to `POST` to `/auth/provider` would pose a problem in cases where the user wants to access a protected page, and the app redirects the user (which can only be followed by a `GET`) to login page. It seems all those redirects would break with a 404 `No route matches [GET] "/auth/provider"` error.

This is essentially the problem we are having - trying to find a clean workaround.  The only idea we had is an intermediate page with a POST login button that is submitted with JavaScript.

mathieujobin:
I cannot believe this hasn't been merged, with all this time 

it was a fine patch

mariokam:
Hey, 
seems like someone have added this CVE recently and currently we're having problem with blocked CI due to security checks with bundler audit checking against CVE database, can someone pls review this and publish the fix as soon as possible ? 
Thanks

svpersteve:
We resolved this with [this gem](https://github.com/cookpad/omniauth-rails_csrf_protection) but this morning our CI was blocked by this CVE. 

Removing Omniauth as it recommends isn't the only way to resolve this. I've temporarily disabled CVE checking in our CI builds so we can continue work

sikachu:
Ah yes, just noticed our CI started to fail with this too, despite we already put in mitigation in place:

```
Updated ruby-advisory-db
ruby-advisory-db: 381 advisories
Name: omniauth
Version: 1.9.0
Advisory: CVE-2015-9284
Criticality: High
URL: https://github.com/omniauth/omniauth/pull/809
Title: CSRF vulnerability in OmniAuth's request phase
Solution: remove or disable this gem until a patch is available!

Vulnerabilities found!
Exited with code 1
```

_If you know what you are doing_ and have mitigated this attack by validating the request source to the request phrase you can ignore the message by passing `--ignore` argument to `bundle audit check` like this:

```console
$ bundle audit check --ignore CVE-2015-9284
```

sikachu:
Small update: I've [proposed a feature in ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db/pull/390#issuecomment-509108186) to let developers mark a CVE as mitigated/fixed so they can mark this issue as fixed in their repository once they implement a fix.

Envek:
Moreover, this vulnerability is application-specific. Only applications that implement authentication via multiple identity providers to the single application's account are vulnerable.

I believe that our application isn't vulnerable, because we don't have any “our own” accounts, we don't have logins and passwords, and we implement only one single identity provider. So we have only “Sign in via eBay” button and nothing more as we're just a convenient tool for eBay users.

> In this scenario, the GitLab account would be linked to the attacker's Google account

In our case it is just impossible because if someone sign in to our app with another eBay account, we will authenticate them to another internal account in our app (or will create brand new one). Because we have strict 1-1 link between eBay accounts and our records in database.

grosser:
FYI Summary of how to mitigate:
 - use `bundle audit check --ignore CVE-2015-9284`
 - add `gem 'omniauth-rails_csrf_protection'` (make sure it's in the default required group)
 - change omniauth links to `method: :post`
 - for each change add `# TODO: remove once https://github.com/omniauth/omniauth/pull/809 is resolved`

jrochkind:
I remain concerned that a developer coming to this repo without having been aware of this whole thing has no instructions in the README warning them about the vulnerability or telling them how to resolve it. 

Does anyone have a sense of what the proper warning/instructions would include, keeping in mind it may need to address Rails usage as well as non-Rails usage?

pat:
Hi all,

Not that more opinions are really needed here - this thread's quite epic - but I'm just pondering what the shorter-term and longer-term resolutions for this might be. Also, folks have done some solid work in remedying this in specific apps, and maybe that can be combined.

In the short-term (e.g. a minor- or patch-level release): can we take the `rack-protection` solution that @abrom's put together, pull it into this gem and make the default behaviour for any POST requests?
https://gist.github.com/abrom/effe58b27a4f4ac1b97fb85593d7c3be

On top of that, would it be feasible to have some conditional logic using @sikachu's solution for Rails used as an alternative, if it's clear that OmniAuth is being used in a Rails context? (and/or don't try to use `rack-protection` if `omniauth-rails_csrf_protection`). Yes, it'd be nice if Rails logic didn't exist here, but it sounds like that's going to need a significant enough change for OmniAuth v2. Potentially this could all be wrapped in an optional setting, so it can be enabled by choice.

I feel like something like this could help with resolving the outstanding CVE, without needing a v2 release - even if the fixed approach isn't enabled by default.

Speaking of v2, that feels like the longer term solution:

* enable the combined approach that I've just suggested as default behaviour.
* default to only POST in the allowed request methods.
* Direct people to use either `omniauth-rails_csrf_protection` or `omniauth-rails` (which I presume would take the same approach).

And for both releases, get the README up-to-date as suggested by @jrochkind. But once we've got the shorter term fix, at least that removes the stress of rushing v2 out the door.

I've got a touch more spare time at the moment, so I'm willing to dive into this if maintainers would like such an approach… but I've no idea if what I'm suggesting is feasible and correct (particularly keen on input from maintainers, @sikachu, @abrom - or anyone else really 😄)

pat:
Further to my comment yesterday:

* I've written a minimal PR #963 that allows `rack-protection` to be used to secure non-Rails applications (essentially, the approach suggested by @abrom, with some tests and clear 403 responses). It does not change existing behaviour in the gem, just provides a slightly smoother path for those dealing with non-Rails apps - essentially, a companion to `omniauth-rails_csrf_protection`.
* I've added a page to the wiki which I feel condenses the wisdom and solutions people have suggested here to resolve the CVE: https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284

Thoughts on both are very welcome - I've marked the wiki page as a draft for now, because getting feedback on it is, I feel, essential before people start using it as a definitive source. But once folks are happy with it, it could potentially be linked to prominently in the README?

zeeshangulzar:
@grosser @pat Do we also need to change anything at stripe dashboard? Or transition from get to post request type will be handled by default?

hiroshi:
Hi, I wrote a couple of tests for a rails app to understand what's going on myself and prevent possible accidental regressions.

I hope it can be help.

They fail without [omniauth-rails_csrf_protection gem](https://github.com/cookpad/omniauth-rails_csrf_protection).

Of course, you may replace `google_oauth2` with your choice of provider.

```ruby
# Make sure that https://nvd.nist.gov/vuln/detail/CVE-2015-9284 is mitigated
RSpec.describe 'CVE-2015-9284', type: :request do
  describe 'GET /auth/:provider' do
    it do
      get '/auth/google_oauth2'
      expect(response).not_to have_http_status(:redirect)
    end
  end

  describe 'POST /auth/:provider without CSRF token' do
    before do
      @allow_forgery_protection = ActionController::Base.allow_forgery_protection
      ActionController::Base.allow_forgery_protection = true
    end

    it do
      expect {
        post '/auth/google_oauth2'
      }.to raise_error(ActionController::InvalidAuthenticityToken)
    end

    after do
      ActionController::Base.allow_forgery_protection = @allow_forgery_protection
    end
  end
end
```

agraves:
Is there a set of blessed steps to remediate this CVE? It's not clear to me which of the suggestions in here are the authoritative way to remediate the CVE github is sending me. Appreciate it

CHTJonas:
> Is there a set of blessed steps to remediate this CVE?

My understanding is that you can mitigate this vulnerability (assuming you're running Rails) with the following steps:

1. Add `gem 'omniauth-rails_csrf_protection', '~> 0.1.2'` to your Gemfile.
2. Change all OmniAuth links to use the POST method eg. `link_to 'Login with Google', '/auth/google_oauth2', method: :post` or use `button_to` instead of `link_to`.

In addition, you may need to use `bundle audit check --ignore CVE-2015-9284` if you are running Bundler audits. You could also consider adding some testing such as the [Rspec example](https://github.com/omniauth/omniauth/pull/809#issuecomment-512689882) given by @hiroshi or an equivalent [MiniTest example](https://gist.github.com/CHTJonas/70cd9ec5fcffa6ca5bae0e04ec51d174).

If you're running Sinatra then you may want to check out the code [here](https://gist.github.com/abrom/effe58b27a4f4ac1b97fb85593d7c3be) by @abrom.

agraves:
@CHTJonas thanks my dude

agraves:
I've taken the liberty of adding this to the [FAQ](https://github.com/omniauth/omniauth/pull/809). If anyone sees errors please make corrections as you see fit.

Not sure if there are other avenues where we should post this information but I suspect many developers are getting notified about this issue and not remediating because the steps are unclear.

a544jh:
I'd like to question if this is really an issue with the omniauth gem itself at all.
I'm not very familiar with Ruby or Rails, so please correct me if I make any wrong assumptions.

To my understanding omniauth is a low-level library to provide an easy way for Ruby web applications to integrate with different authentication providers.
In the reported vulnerability, it is claimed that it can be used to link a primary account to a secondary account owned by an attacker. **To my understanding, omniauth does not provide any session management or "account linking" features at all**, those are to be handled by the caller. Omniauth only provides identity information gathered from the authentication provider.

The README explains something about Rails integration, but I don't really understand how it works. If would understand if Rails has some implicit session handling that could automatically "link" two identities, but I doubt this is the case. Even if that would be the case, it would still be an issue with Rails and not omniauth, since Rails would perform the "linking" action without user feedback, which I see as the actual vulnerability here.

Therefore I think that the CSRF mitigation is irrelevant, since the vulnerability is caused by the user of omniauth, and not omniauth itself.

josh-m-sharpe:
@pat Thanks for consolidating the findings from this over here:

https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284

In the Rails section, it looks like you can drop the 2nd recommendation about only allowing post requests since `omniauth-rails_csrf_protection` handles that for us [here](https://github.com/cookpad/omniauth-rails_csrf_protection/blob/master/lib/omniauth/rails_csrf_protection/railtie.rb)

Quick question though.  I still have a get request for the failure route like so:
`get "/auth/failure" => "omniauth#rejected"`

What is the impact of disabling get requests for that route?  If I recall correctly that was incredibly difficult to test in the real world.  I think oauth providers redirect to that route on failure?  Do you know if that will continue to function if get requests are entirely disabled?

Envek:
@josh-m-sharpe, failure route is where user is being redirected by identity provider in case when authentication is failed by some reason (usually when user didn't give permission to access all their data by your app). On that page your app should display some relevant error message. 

So failure route will just stop working if you disable GET requests for it. And I don't see any meaning in disabling GET requests there.

pat:
Hi Josh,

Good question about the failure endpoint. My knowledge about this is _far_ from complete, but I'm just looking through the source now…

From what I can see (particularly in `OmniAuth::FailureEndpoint`), the redirect occurs from within OmniAuth - ie. a 'bad' or 'failed' request comes back from the provider, which OmniAuth notices and then redirects to `'/auth/failure'`. So, it's not the provider itself making that redirect request.

There's no saving of credentials happening in that failure endpoint - it's just getting a few pieces of data and sending them through in the redirect. So, the question of whether to change your route from a GET request to something else are entirely up to whatever you're doing in that endpoint. I'd expect it's not modifying any data, and is likely fine, so keeping it as a GET is safe.

If you _are_ changing data in your `rejected` action, I'd recommend reconsidering whether that's the right place to do such things. An alternative is to write your own failure endpoint (replacing `OmniAuth::FailureEndpoint` and set via `OmniAuth.config.on_failure`) for any special behaviour.

But yes, the short answer is what @Envek's already shared while I was writing this: you're probably fine keeping it as a GET request.

(And thanks for the wiki note, I'll update that now 😄)

josh-m-sharpe:
Hmm, I was trying to understand if adding the recommended config: `OmniAuth.config.allowed_request_methods = [:post]` will somehow break the `get "/auth/failure"` route.

pat:
From what I've seen in the code, I don't think the two are related.

matteeyah:
Is there any effort being put into resolving this?

redconfetti:
I'm curious about if this issue is only possibly affecting OAuth providers, and not other authentication providers such as CAS, HTTP Basic, LDAP, Shibboleth, etc?

pat:
@redconfetti I'm not familiar enough with those strategies, but from my understanding this issue is only a significant problem when:

* there are multiple authentication sources
• a single user can be associated with more than one authentication source
* authentication can be prompted via a `GET` request (and thus skipping CSRF verification)

So, if any of these other strategies hit all of these issues, then maybe there is something to fix?

pat:
Just had a conversation with @mhaylock, who (correctly) challenged my previous assumptions. I had misunderstood the initial discussions and the CVE. Upon re-reading, I think that this security issue should still be considered a problem if you're using a single OAuth provider, because they may have a CSRF security hole of their own that allows attackers to hijack an authentication flow into their account on the OAuth provider's site.

So: if you're using OmniAuth with _any_ number of providers I would recommend using the mitigation strategies as previously outlined: https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284 (I've removed the 'only-if-using-multiple-providers clause from that page).

Apologies for the extra complication in resolving this!

gencer:
What if we use OmniAuth as a `Single-Sign-On`? In this case, I redirect client to the correct provider. So this is a `GET` request. Should I manually render a page then `POST` it?

Basically, I am not using any kind of Social platforms to sign in to my platform. I allow my clients to authenticate against my own provider (server-side + client-side, both mine).

redconfetti:
We're using [OmniAuth with CAS](https://github.com/dlindahl/omniauth-cas), which uses a 301 redirect to the external CAS server to authenticate and obtain a token that the user is redirected back with. The application then performs a verification of the token (between just the application and CAS service), then establishes the users's application session if the token is valid.

Many points in our application redirect to `/auth/cas`, which then redirects to the CAS server login page. Implementing a forced POST seems very difficult, and I am not sure if it's necessary for our method.

I'm under the impression that because our institutions CAS server doesn't automatically provide an auth token unless you're already logged in for a specific account, and our application would then establish a session for the account you're authenticated as (the attackers account), that this isn't a hazard for those using OmniAuth as a dependency of OmniAuth-CAS.

emilebosch:
Heya, I am wondering. Is there a single source of information like a clear 20 lines of text on what needs to be done in order to secure rails apps? I'm an okay coder but due to information density it is kind of hard to extract out what needs to be done to mitigate this vuln.  I can sorta get it, but I think a lot of junior people might have an issue with it. Would be nice if we put that link in a CVE description.

What is the reason that we just don't allow GET requests and bump the version to major 😬 👿 so that people can at least get rid of their audit message by making the changes and upgrading?

CHTJonas:
> Is there a single source of information like a clear 20 lines of text on what needs to be done in order to secure rails apps?

Yes! All info you need should be on this page of the wiki: [Resolving CVE 2015 9284](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284).

> Due to information density it is kind of hard to extract out what needs to be done to mitigate this vuln

I sympathise - this is a **monster** thread. I managed to get the GitHub "This page is taking too long to load" fail unicorn several times when attempting to make this reply!

bonsohi:
Eagerly awaiting closure on this! Thanks in Advance 🙏

saghaulor:
Hey all, 

I really hate to muddle this issue any further, but I really don't understand how this is a problem for Omniauth, or Rails. I agree with @a544jh in their comment https://github.com/omniauth/omniauth/pull/809#issuecomment-519076052 The reported issue seems specific to workflows where the callback endpoint will associate the identity from the OP to the authenticated user in the vulnerable app. Granted, a lot of sites do this, in particular sites that allow you to login with multiple social identities, but it's not a generic vulnerability that affects everyone. The CSRF vulnerability is only any issue for apps that do this aforementioned workflow. If on the other hand, you are using OIDC to do a find_or_create type flow (probably a single OP scenario) using the claims about the user returned from the OP, this account takeover via CSRF vulnerability doesn't affect you.

There seems to be a lack of clarity on who this issue affects, and because it can lead to some pretty serious consequences (account takeover), people are rightfully excited to get this fixed. But I don't think that this applies to everyone. I think people need to understand clearly if it's a problem that affects them or not, because this isn't a problem with Omniauth or Rails, it's a problem with any app that is an OIDC/Oauth2 consumer that allows identities from multiple OPs to associate to an authenticated user via an endpoint that responds to GET requests.

If I'm mistaken on my assessment, I'm happy to receive feedback. I'd also like to thank everyone involved in trying to address this issue for those who would be affected.

jceaser:
doing a `bundle audit check --ignore CVE-2015-9284` is not good enough for those organizations that have automatic security scans in place. I for one am going to continue to get pinged on using this gem until the CVE goes away despite the fact that we have taken the recommended steps to correct this. Please consider that and come up with a solution that makes the error go away.

amplicity:
Hi all, is the gem below the recommended upgrade route for those of us using Rails?

https://rubygems.org/gems/omniauth-rails_csrf_protection

hiroshi:
> Hi all, is the gem below the recommended upgrade route for those of us using Rails?
https://rubygems.org/gems/omniauth-rails_csrf_protection

Hi, I found that README is updated recently.
https://github.com/omniauth/omniauth#integrating-omniauth-into-your-application
> **Please note:** there is currently a CSRF vulnerability which affects OmniAuth (designated [CVE-2015-9284](https://nvd.nist.gov/vuln/detail/CVE-2015-9284)) that requires mitigation at the application level. More details on how to do this can be found on the [Wiki](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284).

In the wiki page, `omniauth-rails_csrf_protection` is used as a first step for mitigating the vulnerability.

BTW, depedabot may urge to update omniauth to 1.9.1, but I don't think the update don't include any fix for the vulnerability.

See https://github.com/omniauth/omniauth/compare/v1.9.0...v1.9.1
Security related commits seem to be only this REAME change;
https://github.com/omniauth/omniauth/commit/0740ac90614228402952c20a6a3f6ece6c25c57d
, and it only mentioned that `there is currently a CSRF vulnerability that affects OmniAuth` in README.

BobbyMcWho:
@hiroshi this is correct, unfortunately a resolution for that CVE was not included in that version bump. I am not sure of any time-frame in which it will be resolved at this point in time.

DouweM:
> EDIT: @DouweM, update the OP with those information if you please.

@hiroshi Done; thanks for the ping!

tibbon:
What's the status on this one?

josh-m-sharpe:
At this point, is there any reason we can't:

1. munge the learnings from `omniauth-rails_csrf_protection` into this `omniauth`
2. add something like `raise 'this version is only compatible with rails, read [this wiki] for more info and submit a PR! :)' unless defined?(Rails)` somewhere
3. release 2.0 with notes about the breaking changes and how to fix `link_to`
4. wait for the patches for <insert your favorite framework> to come a rollin' on in?

This would solve the issue for (presumably) the majority of us, would be a step in the right direction since the rails solution is more or less figured out and would open the door to other frameworks to contribute an acceptable solution for them.  There's already good work done above for sinatra and rack apps.  They just need to be finalized.

I realize this isn't ideal, but it wouldn't outright break non-rails apps since the `raise` statement prevents them from upgrading.   It might even spur some benefit because it would force the non-rails community to think through what their solution is and get it in place.

We've had some time to let the dust settle, and now I fear that this lingering CVE is conditionalizing everyone to accept that "open CVEs for omniauth" are quite alright.

janko:
I'm trying to take this vulnerability seriously, but I just can't help but think how *difficult* it would be for an attacker to exploit this vulnerability in the way described by the OP.

- the attacker can only target victims that are using a browser session where the attacker is currently logged into a 3rd party service (e.g. Google), which drastically narrows down the victim pool
- the victim needs to be logged into the particular application the attacker wants to hijack (e.g. GitLab.com)
- the attacker needs to feed the victim a link to the `/auth/:provider` page and get them to click on it
- when the 3rd party redirects back, and the app links attacker's 3rd party to victim's primary account, the app would probably show a message such as  "Your Google account has been successfully linked to your primary account", which the victim would have to ignore and not try to unlink it or anything (remember, they didn't intend to sign into GitLab.com or even visit it in the first place, they clicked on a disguised link)

Is there an angle that I'm missing? Or is the main pain of this vulnerability the fact that it has a CVE assigned that is causing noise.

I'm currently building OmniAuth integation for Rodauth, and I thought I should go ahead and disallow `GET /auth/:provider. But from reading a bit, requiring `POST` appears to have its own issues as well, such as being incompatible with authentication flows where something redirects to `/auth/:provider` (which multiple people mentioned here), and not supporting `env["omniauth.params"]` out of the box.

BobbyMcWho:
I want to say the fact that we don't persist query params from the callback phase isn't due to the POST method, but the fact that we set the `omniauth.params` to try and delete the [same key from the session](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb#L236) which I think has to do with the fact that [callback urls contain sensitive information](https://www.oauth.com/oauth2-servers/redirect-uris/). If you really wanted to access the callback query params I _believe_ they should always be accessible on a `Rack::Request.new(@env).GET` since as far as I can tell the [GET](https://github.com/rack/rack/blob/d3225f7c201320ed272a2fa7b000c5850e4a5f88/lib/rack/request.rb#L426-L434) and [POST](https://github.com/rack/rack/blob/d3225f7c201320ed272a2fa7b000c5850e4a5f88/lib/rack/request.rb#L440-L463) methods are available on a Rack::Request regardless of the fact that the HTTP_METHOD was POST. You could do this in a `before_callback_phase` block ~pseudo: 
```ruby
OmniAuth.config.before_callback_phase = ->(env) { env['rack.session']['my.params'] =  Rack::Request.new(env).GET.slice("keys", "you", "want") }
```

janko:
@BobbyMcWho `omniauth.params` isn't for retrieving callback phase parameters, but *request* phase parameters. Supporting POST parameters should be as easy as using `request.params` instead of `request.GET` on [this line](https://github.com/omniauth/omniauth/blob/c95a5a0a50e012b3ba34401a19d589fcdf02afdf/lib/omniauth/strategy.rb#L208).

BobbyMcWho:
It seems that was intentional, changed in [this PR](https://github.com/omniauth/omniauth/pull/867) and [this(https://github.com/advisories/GHSA-9pr6-grf4-x2fr) CVE.

BobbyMcWho:
I'm open for comments/review on this PR: https://github.com/omniauth/omniauth/pull/1010

urkle:
For the use-case where I have my own controller actions that do work before redirecting to /auth, is there a way to simply "call" the request phase manually from my controller? (vs redirecting).. Then I don't even need the /auth/:provider routes at all.

e.g.  our flow is this.
1) Connect integration "links" to a POST to our oauth controller
2) our oauth controller
  a) does validation
  b) loads up the "integration" they are requesting
  c) Sets up the session with details about what the user is trying to do
  c) finally redirects to the "/auth/:provider" endpoint
3) omniauth processes setup + request phases and redirects to provider's endpoint

A more ideal approach would be doing something like this.
1) Connect integration "links" to a POST to our oauth controller
2) our oauth controller
  a) does validation
  b) loads up the "integration" they are requesting
  c) Sets up the session with details about what the user is trying to do
  d) calls the setup + request phases in omni auth that responds with the redirect that needs to occur (going to providers endpoint)

BobbyMcWho:
> For the use-case where I have my own controller actions that do work before redirecting to /auth, is there a way to simply "call" the request phase manually from my controller? (vs redirecting).. Then I don't even need the /auth/:provider routes at all.
> 
> e.g. our flow is this.
> 
>     1. Connect integration "links" to a POST to our oauth controller
> 
>     2. our oauth controller
>        a) does validation
>        b) loads up the "integration" they are requesting
>        c) Sets up the session with details about what the user is trying to do
>        c) finally redirects to the "/auth/:provider" endpoint
> 
>     3. omniauth processes setup + request phases and redirects to provider's endpoint
> 
> 
> A more ideal approach would be doing something like this.
> 
>     1. Connect integration "links" to a POST to our oauth controller
> 
>     2. our oauth controller
>        a) does validation
>        b) loads up the "integration" they are requesting
>        c) Sets up the session with details about what the user is trying to do
>        d) calls the setup + request phases in omni auth that responds with the redirect that needs to occur (going to providers endpoint)

If this is a specific question about the PR I posted, please ask it on the PR  https://github.com/omniauth/omniauth/pull/1010 itself. If this is a general OmniAuth question, please refrain from asking in this thread since there are so many people subscribed to it.

BobbyMcWho:
I've opened a discussion on a v2.0.0 release candidate: https://github.com/omniauth/omniauth/discussions/1017

BobbyMcWho:
The vulnerable by default configuration has been removed in the [v2.0.0 release](https://github.com/omniauth/omniauth/releases/tag/v2.0.0)

PR #818, opened 2015-08-03
Use `session` method instead of `@env['rack.session']`

yui-knk:


nicoolas25:
It will be very nice for me to configure Omniauth un a way that allows me to use another session key, for instance `rack.production.session` instead of `rack.session`

yui-knk:
If you want to use `rack.production.session` as session key, you can override `#session` method.
And this PR make it easy to override session key :)

nicoolas25:
I just added #819 about this. It includes your patch.

nicoolas25:
Can you make the `rack.session` string a frozen constant?

PR #819, opened 2015-08-04
Add the config.session_key to override 'rack.session'

nicoolas25:
Hello,

This PR allows this kind of configuration:

``` ruby
OmniAuth.config.session_key = "rack.production.session"
```

Maybe some of you will find it usefull as I do ^^

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -170,7 +170,7 @@ module OmniAuth
     #
     # @param env [Hash] The Rack environment.
     def call!(env) # rubocop:disable CyclomaticComplexity, PerceivedComplexity
-      unless env['rack.session']
+      unless env[OmniAuth.config.session_key]
         error = OmniAuth::NoSessionError.new('You must provide a session to use OmniAuth.')
         fail(error)
       end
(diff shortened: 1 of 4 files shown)


lib/omniauth/strategy.rb
```diff
@@ -208,9 +208,9 @@ def request_call # rubocop:disable CyclomaticComplexity, MethodLength, Perceived
         call_app!
       else
         if request.params['origin']
-          env['rack.session']['omniauth.origin'] = request.params['origin']
+          session['omniauth.origin'] = request.params['origin']
```
yui-knk: This is not instance variable in origin code, I feel maybe there are some reason to not use instance variable (but author only knows...).

nicoolas25: It isn't nope. It seems to use the default attribute accessor so it basically does the same thing. I updated it for clarity and it does not break the test suite, that's good enough for me ^^

nicoolas25:
Since Rack isn't yet allowing you to update the `env`'s key where the session is stored (`rack.session`) this have a very small use case.

I'll close it but I think the cleaning in #818 may be worth merging.

pas256:
I don't understand how to use this?
How do I inject something into the Rack env so that I can then use this?
For background, what I am doing is creating the OAuth redirects in my Rails 5 API only backend, so there is no session.

PR #820, opened 2015-08-21
Loosen the rack requirement to allow for testing with upcoming rack 2

twalpole:
Rails master has just moved to requiring the rack master branch which is working on rack 2.0 - This loosens up omniauths requirement to allow for testing and development. I specifically need it for testing devise with rails master

### DIFF
--- a/lib/omniauth/builder.rb
+++ b/lib/omniauth/builder.rb
@@ -12,7 +12,7 @@ module OmniAuth
     end
 
     def rack14?
-      Rack.release.split('.')[1].to_i >= 4
+      Rack.release.start_with?('1.') && (Rack.release.split('.')[1].to_i >= 4)
     end
 
     def rack2?
--- a/lib/omniauth/builder.rb
+++ b/lib/omniauth/builder.rb
@@ -2,7 +2,7 @@ module OmniAuth
   class Builder < ::Rack::Builder
     def initialize(app, &block)
       @options = nil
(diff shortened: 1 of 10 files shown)


pwnall:
This is biting me too. Is there anything that I can do to help get this PR in? 

Is any part of it concerning? If so, I could try to come up with a different approach.

Thank you!

lib/omniauth/builder.rb
```diff
@@ -15,6 +15,10 @@ def rack14?
       Rack.release.split('.')[1].to_i >= 4
```
md5: It looks like this should be tightened up to check that the first component of the version number is `1`.

twalpole:
@md5 ok I fixed that check - 4 months ago when I created this PR all the tests passed correctly, now all the ruby 1.8.7 tests fail.  Obviously some gem has been updated to only support ruby 1.9.3+ and the gem restrictions in this project need to be updated if 1.8.7 support is still considered needed.  That being said, there has been no official response to this PR yet so I'm reticent to put in more work figuring out which gem(s) are the issue until there is at least some kind of response

md5:
@twalpole I totally understand regarding the Ruby 1.8.7 thing. I ran into it with #829 as well.

sferik:
Thanks!

md5:
:+1:

pwnall:
Thank you!

PR #826, opened 2015-11-03
"Different callback_url" bug fix

gioblu:
Removing the code and other paramters it works.


tmilewski:
Closing for now as this fix is a little too ambiguous. Would you mind providing more information as well as specs, should this still be problematic?

[Issue Template](https://github.com/omniauth/omniauth/blob/master/.github/ISSUE_TEMPLATE.md)

PR #828, opened 2015-11-11
Stop mutating given hash in #add_mock

fujimura:
`#add_mock` was stringifying original auth hash.


sferik:
Thanks @fujimura!

olivierlacan:
Just got bit by this in 1.1.4. Thanks @fujimura.

Issue #823, opened 2015-09-30
Consider cloning @options when OmniAuth::Strategy is dup'd


md5:
I ran into an issue yesterday with `omniauth-saml` where the `options` data structure was being mutated in `request_phase` in a way that making state bleed across requests (see https://github.com/PracticallyGreen/omniauth-saml/issues/54).

After some initial confusion on my part about the middleware lifecycle and looking through a bunch of code, I found that OmniAuth is doing `dup` before it calls the strategy methods (cf. https://github.com/intridea/omniauth/blob/c3e3aa5bbfb589db1ebed40324c7eea2e3153f8c/lib/omniauth/strategy.rb#L164).

The issue I want to raise here is whether or not it makes sense for `OmniAuth::Strategy` to implement `initialize_copy` like so:

``` ruby
def initialize_copy(orig)
  super
  @options = @options.clone
end
```

This would make it safe for strategy subclasses to mutate `options` in `request_phase`, etc. It might actually need to be `@options.deep_dup` instead of `@options.clone`, but you get the idea.

Does this make sense as a PR for OmniAuth or should I stick with making a PR for `omniauth-saml` that avoids mutating the shared instance of `OmniAuth::Strategy::Options`? If this change to OmniAuth isn't desirable, perhaps it should be documented that it isn't safe for subclasses of `OmniAuth::Strategy` to mutate `options` without making a local copy first.

md5:
If the documentation route is preferable, I can add some text to the [Strategy Contribution Guide](https://github.com/intridea/omniauth/wiki/Strategy-Contribution-Guide) on the wiki.

md5:
I ended up submitting a PR to `omniauth-saml`: https://github.com/PracticallyGreen/omniauth-saml/pull/56

If there's a favorable response to my suggestion here, I can close that one and make a similar PR for `omniauth` itself.

md5:
Bumping this issue to see if anyone has any opinions about implementing a generalized fix for this issue in `omniauth` itself instead of expecting individual strategies to do it

md5:
This was fixed in #829

PR #829, opened 2015-11-12
Duplicate options when a strategy is duplicated

md5:
This PR adds an implementation of `initialize_copy` to `OmniAuth::Strategy` that calls `@options.dup`. Since `@options` is a `Hashie::Mash` instance (:see_no_evil:), calling `dup` does a deep copy.

This addresses the issue that I reported in #823 and https://github.com/PracticallyGreen/omniauth-saml/issues/54.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -479,14 +479,15 @@ module OmniAuth
       OmniAuth.config.on_failure.call(env)
     end
 
+    def dup
+      super.tap do
+        @options = @options.dup
+      end
+    end
+
     class Options < Hashie::Mash; end
 
   protected
-
-    def initialize_copy(orig)
-      super
-      @options = @options.dup
-    end
 
     def merge_stack(stack)
       stack.inject({}) do |a, e|
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -483,6 +483,11 @@ module OmniAuth
 
   protected
 
+    def initialize_copy(orig)
+      super
+      @options = @options.dup
+    end
+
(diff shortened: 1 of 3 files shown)


md5:
Looks like there are failures on Ruby 1.8 since `initialize_copy` is only called in Ruby 1.9+. There are also some Coveralls errors and warnings from Rubocop making other builds fail...

md5:
I cleaned up the Rubocop failures introduced by my original diff and tried to fix the Ruby 1.8.7 failure by moving from implementing `initialize_copy` to overriding `dup`, but the Ruby 1.8.7 and jruby-18mode builds are still failing. I tried unsuccessfully to get `rbenv` to install `1.8.7-p375` to test that version locally, so I'm not able to easily figure out what's going on here...

md5:
Looking at some of the other recent Travis builds for 1.8.7 (e.g. https://travis-ci.org/intridea/omniauth/jobs/88984442), it doesn't appear that these failures were introduced by my changes.

md5:
I just noticed that `rubocop` is not being run for Ruby 2.2, otherwise those build would be failing as well.

It also seems that if 1.8.7 still supported, then `Style/TrailingComma` should not be set to `comma` in `.rubocop.yml`.

sferik:
Thanks @md5!

md5:
It's great to see this merged! Do you have a timeline for cutting a new release by any chance?

I see you've got the Travis tests passing, so hopefully a release is coming soon. :+1:

sferik:
Released [version 1.3.0](https://rubygems.org/gems/omniauth/versions/1.3.0).

md5:
:metal:

PR #831, opened 2015-12-19
Hashie >= 2.1 doesn't support Ruby 1.8 (CI build fix! 🍣🍣🍣=͟͟͞͞)

amatsuda:
We see so many CI failures against ruby 1.8.7 in the PRs on this repo, and due to which, I suppose, this repo seems somewhat inactive recently.

This problem is actually totally unrelated to changesets in those PRs, but it's because we're bundling Hashie 3.x that doesn't support Ruby 1.8.

After merging this, we would be able to merge other good ones such as #820, #828, #829, etcetc.

see: https://github.com/intridea/hashie/blob/master/CHANGELOG.md#210-462014

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ group :test do
   gem 'coveralls'
   gem 'json', '>= 1.8.1', :platforms => [:jruby, :ruby_18, :ruby_19]
   gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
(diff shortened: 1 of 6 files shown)


amatsuda:
Can any maintainer take a look on this please?
We want you to merge this and #820, then make a gem release in order to let OmniAuth work with Rails 5.
Maybe @sferik?

sferik:
Thanks @amatsuda! I will take a look at the other PRs now.

md5:
:metal:

PR #839, opened 2016-02-28
non-raise-out failure redirects with params from request

kbravi:
Failure endpoint redirects with the params from env['omniauth.params']


BobbyMcWho:
I'm not entirely sure that this is default behavior that we'd want in omniauth, that being said, you can use this behavior in your own app by creating a custom failure handler.

PR #843, opened 2016-03-30
Added check whether ancestor method is a Proc

madmike:
I've got a strange error when updated to rails 4.2.6:
TypeError (wrong argument type Fixnum (expected Proc)) in compile_stack method
When i looked deeper i found out that this method test all the ancestors of the OmniAuth::Strategies::XXX (currently using one, e.g. Facebook). Among them there is sure an Object class which had an ancestor of its own a Process class in production environment. So the Process has a uid method which returns the Fixnum class and caused this error.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,20 @@
+Please complete all sections.
+
+### Configuration
+
+- Provider Gem: `omniauth-*`
+- Ruby Version: ``
+- Framework: ``
+- Platform: ``
+
+### Expected Behavior
+
+Tell us what should happen.
+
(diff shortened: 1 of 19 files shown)


adin-ttc:
I have this issue. I'm sure it has something to do with the way our Environment auto loads but I haven't had time to investigate.. This does solve the problem.

tmilewski:
Is this still an issue? If so, would you mind rebasing on Master?

adin-ttc:
@tmilewski this is indeed still an issue.. I will try a rebase

BobbyMcWho:
If this is still an issue, can you make a clean PR? There's a lot of conflicts presently.

PR #848, opened 2016-05-24
Update AuthHash#regular_writer to ensure that #info is a Hash

estepan:
omniauth-cas is failing with `undefined method #each_pair on String` when unexpected params are passed
https://github.com/intridea/omniauth/blob/master/lib/omniauth/strategy.rb#L336


spec/omniauth/auth_hash_spec.rb
```diff
@@ -8,6 +8,11 @@
     expect(subject.info.first_name).to eq('Awesome')
   end
 
+  it 'does not try to parse `string` as InfoHash' do
```
tmilewski: I'm not convinced that this test covers your use-case as it's technically already tested here:

https://github.com/estepan/omniauth/blob/ed89156612311e0d2ced47c73ea7d0748e5a7eef/spec/omniauth/auth_hash_spec.rb#L5-L9

estepan: `undefined method #each_pair on String` occurs when you try to initialize `Hashie::Mash` with a string.
And it's not tested. Your test describes that it simply parses `info` field which is assigned with a hash as `InfoHash` object. If you assign a string to `subject.info` if would fail. Also `omniauth` tries to parse every `info` on any level of nesting and it causes some troubles, but it's not covered here.

tmilewski: 👍

PR #850, opened 2016-06-22
Avoid ruby warnings

printercu:
Some strategies use minitest, and it has enabled warnings by default. So there are long list  of warnings like this: https://travis-ci.org/mkdynamic/omniauth-facebook/jobs/139413665


tmilewski:
Would you mind rebasing on Master?

printercu:
Rebased

tmilewski:
Thanks! Have you tested that same instance of `omniauth-facebook` with this code and everything works?

printercu:
Yes.

PR #859, opened 2016-09-16
require hashie before hashie/mash

marcqualie:
https://github.com/intridea/hashie/issues/367


dblock:
I released hashie 3.4.6 so this shouldn't be necessary.

fidelisrafael:
So we can close this?

PR #867, opened 2017-01-11
security issue in returning post parameters from session in callback phase

lalithr95:
Request phase of omniauth store `request.params` in session which are later assigned in env of callback phase. According do docs we should only store query params but in this case both GET and POST params get stored. POST params can contain `authenticity_token` of application to protect form CSRF issues. We shouldn't leak such tokens from POST params.

@sferik @jamesarosen @md5

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -198,7 +198,7 @@ module OmniAuth
       setup_phase
       log :info, 'Request phase initiated.'
       # store query params from the request url, extracted in the callback_phase
-      session['omniauth.params'] = request.params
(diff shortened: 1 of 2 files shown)


lalithr95:
ping ? due to security concern

jamesarosen:
I haven't been involved in this project in years and know nothing about this.

sferik:
@lalithr95 Thanks!

lalithr95:
@sferik do we need a minor version bump ?

carnil:
This issue has been assigned CVE-2017-18076 by MITRE.

PR #868, opened 2017-01-12
docs: TestCase expects class in strategy method array

azul:
The documentation included MyStrategy.new which would return an instance unless it fails due to lack of args.


tmilewski:
Would you mind rebasing on Master?

PR #870, opened 2017-01-19
remove security issues from old openssl

tronicum:


tmilewski:
Apologies but I have to close this PR. 

This effectively removes all testing for most Ruby versions and allows failures for those that are left. It's almost as if testing were removed all-together.

PR #874, opened 2017-02-03
Silence Hashie::Mash logger on Hashie 3.5.0+

### LINKED ISSUE #872: Hashie warnings
I receiving the following warning on rails console:
You are setting a key that conflicts with a built-in method OmniAuth::AuthHash::InfoHash#name defined at /home/vagrant/.rvm/gems/ruby-2.4.0@rails5.0/gems/omniauth-1.3.2/lib/omniauth/auth_hash.rb:34. This can cause unexpected behavior when accessing the key via as a property. You can still access the key via the #[] method.

Could some one help me?

michaelherold:
We introduced a new logger for Mash due to the huge number of issues
that we get where people try to set keys on their Mash. This change
silences the info messages that get logged when a Mash has a conflict
with a method name.

I had to do this in a kind of hacky way. Since OmniAuth supports Hashie
versions `>= 1.2, < 4`, I can't just hook into the method that we'll be
adding in 3.5.2. As such, I apply an override for the logging method if
the disable method isn't present.

I'm also only overriding the method in versions that have the logging
capability.

If anyone has a better suggestion on how to handle this, I'm open to
suggestions! I'm also open to changing the `KeyStore` name. I went
with that because `OmniAuth::Mash` wouldn't communicate much if you
aren't familiar with Hashie.

Fixes #872

### DIFF
--- a/lib/omniauth/auth_hash.rb
+++ b/lib/omniauth/auth_hash.rb
@@ -1,11 +1,11 @@
-require 'hashie/mash'
+require 'omniauth/key_store'
 
 module OmniAuth
   # The AuthHash is a normalized schema returned by all OmniAuth
   # strategies. It maps as much user information as the provider
   # is able to provide into the InfoHash (stored as the `'info'`
   # key).
-  class AuthHash < Hashie::Mash
+  class AuthHash < OmniAuth::KeyStore
     def self.subkey_class
       Hashie::Mash
     end
@@ -26,7 +26,7 @@ module OmniAuth
       super
     end
 
-    class InfoHash < Hashie::Mash
+    class InfoHash < OmniAuth::KeyStore
       def self.subkey_class
         Hashie::Mash
       end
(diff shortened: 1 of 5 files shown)


michaelherold:
The test around this are ... convoluted ... because of the requirement for coverage not dropping.

michaelherold:
The failure on 1.8.7 doesn't seem to have anything to do with this change.

tmilewski:
Hey @michaelherold, would you mind either rebasing on master or allowing edits from maintainers? Thanks!

michaelherold:
There you go @tmilewski - I rebased it. I don't see any setting that should be preventing you from editing.

Please let me know if there's anything else I can do!

tmilewski:
Perfect, thank you!

michaelherold:
I just released Hashie v3.5.2 which has the `disable_warnings` feature for the Mash logger.

JulienItard:
@michaelherold 
Hi, how can I use this new `disable_warnings` on rails ? I'm on hashie 3.5.5.
I tryed `Hashie.logger = Logger.new(nil)` in initializer but it doesn't work.

Thanks 👍

PR #889, opened 2017-02-18
Revert #806

tmilewski:


jcohenho:
Why was this reverted? Having headers accessible in the omniauth env vars is very important, especially for mobile authentication.

PR #907, opened 2017-08-10
Add support for using developer strategy behind mapping

GameDevFox:
If the Developer strategy is used behind middleware like this:

```
map '/api'
  use OmniAuth::Strategies::Developer

  # callback here for /auth/developer/callback
end
```

the callback path that is used in the form does not contain the
ENV['SCRIPT_NAME'] before the callback_path

Currently the form path would be:
/auth/developer/callback

But since our callback handler is also behind this mapping it
should be:
/api/auth/developer/callback

This fix prepends `script_name` to the `callback_path` to give us
the problem form submission url regardless of whether it's behind
a mapping or not

### DIFF
--- a/lib/omniauth/strategies/developer.rb
+++ b/lib/omniauth/strategies/developer.rb
@@ -35,12 +35,16 @@ module OmniAuth
       option :uid_field, :email
 
       def request_phase
-        form = OmniAuth::Form.new(:title => 'User Info', :url => callback_path)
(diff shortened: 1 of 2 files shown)


BobbyMcWho:
You can set the callback_path option: https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb#L397

Issue #910, opened 2017-09-25
attack against user session with omniauth.origin


97jaz:
Please complete all sections.

### Configuration

- Provider Gem: `omniauth-auth0`
- Ruby Version: `2.4.2`
- Framework: `rails-5.1.4`
- Platform: `OS X`, `Linux`

### Expected Behavior

Omniauth shouldn't allow untrusted parties to populate the user session with arbitrary data by blindly trusting the contents of the `origin` query parameter in a `GET` to `/auth/<provider>`.

### Actual Behavior

Omniauth puts the contents of the `origin` query parameter into the [user's session](https://github.com/omniauth/omniauth/blob/df95e5c571e8cd972023b0482ee73752f412b618/lib/omniauth/strategy.rb#L213).

If the application is using cookies for session storage, it is trivially easy to overflow a user's session cookie by redirecting them to `/auth/<provider>?origin=<string longer than 4k>`. (Or, use a value that gets the session cookie close to, but not over, the 4k limit, so that auth succeeds, but any attempt to put anything into the user's session after that causes it to overflow.)

If the application is using different session storage, it opens a possible DOS attack against the storage layer.

### Steps to Reproduce

Navigate to `/auth/<provider>?origin=<long string>`.

--

Maybe I've missed some important aspect of how this feature works, but it seems to me that saving the user's location is much easier to do securely in the application layer, rather than the middleware layer.

tmilewski:
Unless there's something that I'm missing, this check is typically handled through `Rack::Session::Cookie` or, in your case, `ActionDispatch::Cookies::CookieJar`.

**Rack::Session::Cookie:**
- https://github.com/rack/rack/blob/master/lib/rack/session/cookie.rb#L156-L170

**ActionDispatch::Cookies::CookieJar**

-
  https://github.com/rails/rails/blob/1fa268bfa5667a0e9ddbfda243b5282c023ab9ad/actionpack/lib/action_dispatch/middleware/cookies.rb#L589-L593
-
  https://github.com/rails/rails/blob/1fa268bfa5667a0e9ddbfda243b5282c023ab9ad/actionpack/lib/action_dispatch/middleware/cookies.rb#L632-L636

97jaz:
[...]

PR #914, opened 2017-10-02
Adds config option to disable omniauth.params

97jaz:
Similar to #912, this PR adds an option, `store_params`, which controls whether `omniauth.params` is preserved in the session. It defaults to `true` for backward compatibility.

Background: I had a test case in my application for the scenario described in #910, and it continued to fail after updating my version of the omniauth gem to 1.7.0 and 1.7.1. The test sends a very long string in the `origin` request parameter, and expects that it will not result in a cookie overflow exception -- but the exception was still being raised. And that's because OmniAuth was still trying to put the the very long query parameter in the session, this time under `omniauth.params`.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -15,6 +15,7 @@ module OmniAuth
         option :setup, false
         option :skip_info, false
         option :origin_param, 'origin'
+        option :store_params, true
       end
     end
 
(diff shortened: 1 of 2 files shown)


BobbyMcWho:
I'd rather have a configurable option to pick specific param keys in an allow_list, and all others would be filtered out. I'm going through and closing stale PRs, but if you're willing to implement the allow_list instead, feel free to submit a new PR.

PR #917, opened 2017-10-05
Update for frozen-string-literal friendliness.

pat:
One minor change that keeps all tests green when frozen string literals are in play in MRI 2.4+.

I've done this testing by adding the pragma comment to all Ruby files in lib and spec locally, but have not included those changes in this PR. I'm happy to add them in if you'd like, I just don't want to presume whether that's preferred or not.

### DIFF
--- a/lib/omniauth/form.rb
+++ b/lib/omniauth/form.rb
@@ -9,7 +9,7 @@ module OmniAuth
       options[:header_info] ||= ''
       self.options = options
 
-      @html = ''.dup
+      @html = +'' # unary + string allows it to be mutable if strings are frozen
       @with_custom_button = false
       @footer = nil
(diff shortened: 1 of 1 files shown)


lib/omniauth/form.rb
```diff
@@ -9,7 +9,7 @@ def initialize(options = {})
       options[:header_info] ||= ''
       self.options = options
 
-      @html = ''
+      @html = ''.dup
```
BobbyMcWho, line 12:
```suggestion
      @html = +'' # unary + string allows it to be mutable if strings are frozen
```

BobbyMcWho (approved):

BobbyMcWho:
Re-trigger CI

PR #935, opened 2018-05-09
Add before_other_phase hook

sj26:
Add a before_other_phase hook alongside the similar hooks for options, request, and callback phases.


sj26:
Well this got ignored. :-(

PR #948, opened 2018-10-08
Update ruby minor versions for ci

furuhama:
Omniauth ci step runs with old minor version Rubys now.

And, it would be better to update Ruby minor versions for travis ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ language: ruby
 rvm:
   - jruby-9000
   - 2.1.10 # EOL Soon
-  - 2.2.6
-  - 2.3.3
-  - 2.4.0
-  - 2.5.0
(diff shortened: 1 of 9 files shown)


furuhama:
(After all test has passed, I would also update README.)

furuhama:
@tmilewski 
How do you think of this cop?
```
omniauth.gemspec:21:32: C: Gemspec/RequiredRubyVersion: required_ruby_version (2.1, declared in omniauth.gemspec) and TargetRubyVersion (2.2, which may be specified in .rubocop.yml) should be equal.
  spec.required_ruby_version = '>= 2.1.9'
                               ^^^^^^^^^^
```

(Or, when do you plan to drop support for ruby 2.1 version?)

PR #949, opened 2018-12-12
Change debugging logging to use debug channel

alexblackie:
These messages are very specific to omniauth internals, and thus are
only useful when debugging or developing. Having these messages on
`:info` causes them to fill test suite output or other non-production
logs which are at the `:info` level. We should move these messages to
`:debug` so developers can still see them, but they do not pollute
non-development logs.


vfonic:
Indeed, I just came to this repo to figure out how to disable these logs. :)

Any ideas?

BobbyMcWho (approved):

BobbyMcWho:
Triggering CI run

PR #952, opened 2018-12-30
Remove Gemnasium badge

listrophy:
Gemnasium no longer works, so the badge doesn't tell visitors anything.


BobbyMcWho:
This has been removed in another PR

PR #958, opened 2019-05-10
Update the example to match the defaults used by Rails. 

hectorcorrea:
This prevents the "Can't verify CSRF token authenticity" error when using the gem for the first time with Rails.

The default `developer` strategy uses HTTP POST (https://github.com/omniauth/omniauth/blob/master/lib/omniauth/form.rb#L78) to submit the values and Rails by default attempts to validate these requests. 

The [FAQ in the Wiki](https://github.com/omniauth/omniauth/wiki/FAQ) suggests the changes that I am submitting in this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ steps are necessary for your application. For example, in a Rails app I
 would add a line in my `routes.rb` file like this:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


BobbyMcWho:
A similar PR has been merged recently into the 2_0_indev branch

PR #963, opened 2019-07-10
Add optional rack-protection-based CSRF check.

pat:
This is related to #809 and related discussions around CVE-2015-9284. `OmniAuth::AuthenticityCheck` is available to hook into `rack-protection` (provided that gem is available), and will reject requests that do not come with a CSRF token that matches what's stored in the session. This is useful _only_ in non-Rails applications (whereas otherwise people are better served using the `omniauth-rails_csrf_protection` gem).

To enable this approach, you will need to set up the `before_request_phase` hook:

```ruby
OmniAuth.config.before_request_phase = OmniAuth::AuthenticityCheck.new
```

… and then ensure your session's `csrf` token is passed through in each POST request to `/auth/:provider`, as per [the documentation](https://github.com/sinatra/sinatra/blob/master/rack-protection/lib/rack/protection/authenticity_token.rb) in `rack-protection`.


BobbyMcWho:
closed via #1010

PR #964, opened 2019-07-16
Fixing the build on Travis CI

pat:
A bunch of small changes - which I can wrap up into a single commit or two if preferred. Each commit's message is hopefully reasonably clear, but a quick summary:

* Use the latest patch releases of Ruby versions. This also helped the code coverage stats be more consistent in JRuby.
* Ensure bundler v1 is used (which is more work that I'd like - due to RVM insisting on installing bundler v2 as a default gem for JRuby, and thus can't be uninstalled).
* Avoid upgrading ruby gems during the build - again, not ideal, but MRI 2.2 complains otherwise.
* Lock Rubocop to 0.58.

The goal here was to get the build green again, without any code changes (and the only code change that's been made is removing a `freeze` call on a regular expression, because Rubocop says it's already immutable). Rubocop itself is a very old version, but making the upgrade to a newer one will almost certainly introduce more code changes - and I think that _should_ happen at some point, but not in this PR.

There's also the challenge that the latest Rubocop releases don't support the discontinued MRI 2.2 - at what point should OmniAuth make the same jump as well?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 bundler_args: --without development
 before_install:
-  - gem update --system
-  - gem update bundler
+  - gem install bundler --version 1.17.3
 cache: bundler
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
     - JRUBY_OPTS="$JRUBY_OPTS --debug"
 language: ruby
 rvm:
-  - jruby-9000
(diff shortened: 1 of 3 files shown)


BobbyMcWho:
Thanks for your work on this @pat, #977 Took care of the failing CI, I agree with points you make about potentially removing support for 2.2, but that's a discussion for another time. :)

PR #977, opened 2019-11-11
Allow Hashie version 4.0.x

### LINKED ISSUE #976: Update dependencies to support use of hashie 4.0.0?
### Configuration

- Provider Gem: `omniauth-azure-oauth2`
- Ruby Version: `2.6.5p114`

    omniauth-azure-oauth2 was resolved to 0.0.10, which depends on
      omniauth (~> 1.0) was resolved to 1.9.0, which depends on
        hashie (>= 3.4.6, < 3.7.0)

BobbyMcWho:
https://github.com/intridea/hashie/releases/tag/v4.0.0

We released the 4.0.0 release as a major semver change due to multiple non-destructive hash methods being changed to return Mash subclasses instead of plain ol' Ruby hashes. This shouldn't affect the majority of projects, but versioning changes allow projects to choose whether to upgrade or not. 

The other included changes were the result of some yak-shaving I had to do to get CI passing again since it's been failing for 4 months :( 

closes #976

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 bundler_args: --without development
 before_install:
   - gem update --system
-  - gem update bundler
+  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+  - gem install bundler -v '~> 1.14'
 cache: bundler
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 bundler_args: --without development
 before_install:
   - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
-  - gem install bundler -v '~> 1.14'
+  - gem install bundler -v '1.17.3'
+install:
+ - bundle _1.17.3_ install --jobs=3 --retry=3
 cache: bundler
 env:
   global:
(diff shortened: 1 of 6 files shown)


lib/omniauth/builder.rb
```diff
@@ -1,24 +1,5 @@
...
-    end
-
-    def rack14?
-      Rack.release.start_with?('1.') && (Rack.release.split('.')[1].to_i >= 4)
-    end
-
-    def rack2?
-      Rack.release.start_with? '2.'
-    end
-
```
BobbyMcWho: Now that we've locked rack above 1.6, we don't need any of this anymore.

.travis.yml
```diff
@@ -1,7 +1,9 @@
 bundler_args: --without development
 before_install:
-  - gem update --system
-  - gem update bundler
+  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+  - gem install bundler -v '1.17.3'
+install:
+ - bundle _1.17.3_ install --jobs=3 --retry=3
```
BobbyMcWho: bundler 2 doesn't work on ruby < 2.3

supernova32:
Thanks for the work on this PR @BobbyMcWho and thank you for getting CI to be green again.

Sorry for the delay in reviews :(

dblock:
I think this should not have a maximum version of Hashie at this point. There are no known issues with Hashie 5 yet that would prevent any version of omniauth from working with it, so there's no need to restrict.

PR #981, opened 2019-11-18
simplecov ignore strategy_macros

BobbyMcWho:
This file is just macros for tests that strategies can use. It seems that coverage in this is flakey, and we don't need it.


BobbyMcWho:
@supernova32  Can I get a CR approval to merge this?

Issue #940, opened 2018-08-11
Are OmniAuth strategies thread safe?


jrgns:
I've started adding `require 'rack-freeze'` to the top of my `config.ru` file to assess how thread safe my rack middleware is. The first issue I ran into was with omniauth.

### Configuration

- Provider Gem: `omniauth`
- Ruby Version: `2.4.4`
- Framework: `Rack`
- Platform: `Ubuntu`

`config.ru` file:

```ruby
require 'rack/freeze'
require 'omniauth'
use OmniAuth::Strategies::Developer
app = Proc.new do |env|
    ['200', {'Content-Type' => 'text/html'}, ['A barebones rack app.']]
end
run app
```

### Expected Behavior

A request should be handled as normal. So

```bash
$ curl localhost:9292
A barebones rack app.
```

### Actual Behavior

An exception is thrown:

```
[2018-08-11 11:15:20] INFO  WEBrick 1.3.1
[2018-08-11 11:15:20] INFO  ruby 2.4.4 (2018-03-28) [x86_64-linux-gnu]
[2018-08-11 11:15:20] INFO  WEBrick::HTTPServer#start: pid=29537 port=9292
RuntimeError: can't modify frozen OmniAuth::Strategies::Developer
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/omniauth-1.8.1/lib/omniauth/strategy.rb:491:in `block in dup'
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/omniauth-1.8.1/lib/omniauth/strategy.rb:490:in `tap'
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/omniauth-1.8.1/lib/omniauth/strategy.rb:490:in `dup'
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/omniauth-1.8.1/lib/omniauth/strategy.rb:168:in `call'
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/rack-2.0.5/lib/rack/session/abstract/id.rb:232:in `context'
	/home/jrgns/Code/EagerELK/threadsafe/vendor/bundle/ruby/2.4.0/gems/rack-2.0.5/lib/rack/session/abstract/id.rb:226:in `call'
[...]

PR #983, opened 2019-11-21
Dup options when a strategy is dup'd or cloned

ccyrille:
In multi-threaded environments, we should ensure that Rack middlewares are immutable. The [`rack-freeze`](https://github.com/ioquatix/rack-freeze) gem is small a policy framework that ensures just that. More infos are avaialble in [this article](https://crypt.codemancers.com/posts/2018-06-07-frozen-middleware-with-rack-freeze/). 

This keeps the fix introduced in https://github.com/omniauth/omniauth/pull/829 but also ensures that the base strategy is immutable (and pass `rack-freeze` policy). This matches the original implementation which was apparently abandoned for Ruby 1.8.7 compatibility at the time (https://github.com/omniauth/omniauth/pull/829/commits/fa868a0f0d1e879d18c4beb51a905f3aace1e14c).

Related issue : https://github.com/omniauth/omniauth/issues/940

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -491,15 +491,14 @@ module OmniAuth
       OmniAuth.config.on_failure.call(env)
     end
 
-    def dup
-      super.tap do
-        @options = @options.dup
-      end
-    end
-
     class Options < OmniAuth::KeyStore; end
 
   protected
+
+    def initialize_copy(*args)
+      super
+      @options = @options.dup
+    end
 
     def merge_stack(stack)
       stack.inject({}) do |a, e|


BobbyMcWho:
So, I'm going to have to dig into this, 
https://github.com/omniauth/omniauth/pull/829/files was where dup was originally added, and this comment
https://github.com/omniauth/omniauth/pull/829#issuecomment-156187623 mentions that initialize_copy does not work on ruby 1.8. 

I'm personally okay with dropping ruby 1.8 support, but that'll have to come in a major version release, which isn't going to be any time in the near future, as there are some other things I think should be included in a major release.

BobbyMcWho:
I take that back, looking at https://github.com/omniauth/omniauth#supported-ruby-versions , ruby 1.8 is no longer listed. I'll dig in later to see what risk this introduces.

ccyrille:
@bobbymarko Yes, sorry it wasn't perfectly clear that Ruby 1.8.7 didn't have to be supported anymore. For what I see, `initialize_copy` is fully supported since 2.1.10 version (see https://apidock.com/ruby/Object/initialize_copy). 

One more thing to be noticed : I used `initialize_copy` instead of `initialize_dup`, so this will override behaviour both for `dup` and  `clone`. Seemed to me like the right behavior but maybe it isn't, happy to discuss about it.

ccyrille:
@BobbyMcWho Current implementation is buggy by nature because of the strategy mutability and all people using the `omniauth` gem can't ensure their other middlewares are safe by using [`rack-freeze`](https://github.com/ioquatix/rack-freeze). Any news from your side ? Thx !

BobbyMcWho:
Hi @ccyrille, sorry I didn't get to this.

I joined the maintainers for this project hoping to close some issues and help out the existing maintainers. Unfortunately my attempts at communication with the folks with release access to rubygems has fallen flat, and I decided not to put any more effort in if ultimately I wasn't going to be able to release.

ccyrille:
@BobbyMcWho Oh that's a shame ! I have seen that you have already done a great job on the repo 👏 and I perfectly understand that it does not worth the effort if it doesn't reach release stage. I will also try to reach to them to support your effort :)

lib/omniauth/strategy.rb
```diff
@@ -491,16 +491,15 @@ def fail!(message_key, exception = nil)
...
-        @options = @options.dup
-      end
-    end
-
     class Options < OmniAuth::KeyStore; end
 
   protected
 
+    def initialize_copy(*args)
+      super
```
BobbyMcWho, line 499: Do we need to pass args into super here?

PR #991, opened 2020-02-17
Remove Hakiri badge

gjtorikian:
This page provides a really glaring notice on the README: 
image

...except, the latest commits are all warnings/green. The only security issue is from three years ago: 

image

I'm guessing this Hakiri service doesn't really work, considering it also hasn't had an update in three years.


BobbyMcWho:
closing and re-opening to trigger CI

PR #993, opened 2020-02-26
Release 1.9.1

BobbyMcWho:
Have to do this via PR, since the branch is protected.

### DIFF
--- a/lib/omniauth/version.rb
+++ b/lib/omniauth/version.rb
@@ -1,3 +1,3 @@
 module OmniAuth
-  VERSION = '1.9.0'.freeze
+  VERSION = '1.9.1'.freeze
 end
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -10,7 +10,7 @@ if RUBY_VERSION >= '1.9'
   SimpleCov.start do
     add_filter ['/spec/', '/vendor/', 'strategy_macros.rb']
     minimum_coverage(92.5)
-    maximum_coverage_drop(0.01)
+    maximum_coverage_drop(0.05)
   end
 end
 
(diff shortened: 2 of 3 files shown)


supernova32:
Do you know why Travis is failing? It seems odd, though. Is it passing locally?

BobbyMcWho:
It's not passing locally, rack 2.1.0 and 2.2.0 included some changes that broke a few tests.
[It seems they may not really be an issue](https://github.com/rack/rack/issues/1612#issuecomment-591819697) so I just need to update those specs

supernova32:
I think a better option would be to lock down the rack version used for the tests. It doesn't make sense that a simple change like this breaks the build, just because there is a newer version of rack that is not compatible with our code.

Can you lock it down to the version that works? That way master will be green without having to change the tests.

BobbyMcWho:
Becuase rack is a co-dependency of this package and a ton of others, I don't want to artificially limit the version of rack if possible. We should only pin it of it rack actually merges changes that will fundamentally affect the way this package's code functions

supernova32:
Makes sense. But I also don't want to merge this PR when it will lead to a "broken" master. Could you open a new PR with the test fixes, so we merge that one first?

BobbyMcWho:
Yup, I'm at Lonestar Elixir conf currently, so it'll be a few days most likely til I can look at it

supernova32:
Nice! Have fun!

BobbyMcWho:
@supernova32 can you approve this? I think the coverage check is based on master branch, and so I can't get passed it even though I haven't changed anything that should be tested.

techdoutdigi21 (changes requested):
Check check

techdoutdigi21 (changes requested):
.

PR #994, opened 2020-02-28
change get to post in README

meliew:
when following the instructions with the developer provider I noticed that I needed to use post


BobbyMcWho:
Merged in another change that was the same as this, thank you for contributing

Issue #999, opened 2020-04-25
Do we want to memoize #OmniAuth::Strategy#credentials?


zinosama:
### Expected Behavior

`#credentials` to be called once per callback request.

### Actual Behavior

`#credentials` to be called twice per callback request.

### Question

Is there any reason `OmniAuth::Strategy#credentials` is not memoized given that `OmniAuth::Strategy#auth_hash` always invokes it twice in a row:
```ruby
# OmniAuth::Strategy#auth_hash 
hash.credentials = credentials if credentials
```

This results in all providers' `credentials` block to be evaluated twice per callback request.

BobbyMcWho:
I'm open to a PR, I'm not intimate with that section of code in particular

PR #1000, opened 2020-04-25
Remove duplicative eval of credentials and extra

zinosama:
Addresses https://github.com/omniauth/omniauth/issues/999.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -345,11 +345,13 @@ module OmniAuth
     end
 
     def auth_hash
-      hash = AuthHash.new(:provider => name, :uid => uid)
-      hash.info = info unless skip_info?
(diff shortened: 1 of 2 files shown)


lib/omniauth/strategy.rb
```diff
@@ -347,8 +347,8 @@ def extra
     def auth_hash
       hash = AuthHash.new(:provider => name, :uid => uid)
       hash.info = info unless skip_info?
-      hash.credentials = credentials if credentials
-      hash.extra = extra if extra
+      (credentials_data = credentials) && hash.credentials = credentials_data
+      (extra_data = extra) && hash.extra = extra_data
```
zinosama, line 351: Using local variables (instead of instance variables) for memoization to minimize unintended side effect. If you find these one liners to be unreadable, I'm happy to separate assignments into their own lines @BobbyMcWho

PR #1002, opened 2020-05-05
Test against Ruby 2.6.6 and 2.7.1

okuramasafumi:
People use OmniAuth with new Rubys and we need tests for them.


joshRpowell:
ping @BobbyMcWho @supernova32 👀  for a patch release?

PR #1003, opened 2020-05-31
Please make the implementation work in development

jessitron:
The suggested implementation doesn't work in development. It produces a few errors that are baffling to a newbie.
I'm here with a rails expert (@avdi) and it took the two of us a long time to get this working as far as you intended it to.

The development form issues a POST, not a GET. I don't yet know what the other sites do, but POST is necessary for omniauth to work in development.
And the verify_authentication_token check fails, so somehow this has to be worked around in order to trigger the create action.

Please please, for newbies, I don't know that these are the right changes, but make it so that people new to Rails and Omniauth can run in development mode immediately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ steps are necessary for your application. For example, in a Rails app I
 would add a line in my `routes.rb` file like this:
 
 ```ruby
-get '/auth/:provider/callback', to: 'sessions#create'
+post '/auth/:provider/callback', to: 'sessions#create'
 ```
 
 And I might then have a `SessionsController` with code that looks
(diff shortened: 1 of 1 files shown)


BobbyMcWho:
Hi @jessitron, I'm sorry that the examples don't work in development, I apologize that I took so long to get back to you.

If you're still interested in this, can you let me know what version of rails you were running?

README.md
```diff
@@ -86,14 +86,16 @@ steps are necessary for your application. For example, in a Rails app I
...
-get '/auth/:provider/callback', to: 'sessions#create'
+post '/auth/:provider/callback', to: 'sessions#create'
 ```
 
 And I might then have a `SessionsController` with code that looks
 something like this:
 
 ```ruby
 class SessionsController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :create unless Rails.env.production?
```
BobbyMcWho, line 97:
```suggestion
  # If you're using a strategy that POSTs during callback, you'll need to skip the authenticity token check for the callback action only. 
  skip_before_action :verify_authenticity_token, only: :create
```

BobbyMcWho (approved):

PR #1014, opened 2020-12-10
Frozen string support

BobbyMcWho:


BobbyMcWho:
#917 will fix the broken check

PR #1019, opened 2021-01-05
use ruby 2.4 in matrix

enomotodev:
Ruby 2.4 is no longer officially supported, but omniauth's Ruby support is 2.2 and above, so I've also added tests for Ruby 2.4 to GitHub Actions.


BobbyMcWho:
Ruby 2.4 is EoL and will no longer be tested against as of omniauth 2.0

PR #1027, opened 2021-01-16
Allow OmniAuthAuthenticityTokenProtection options to be configured

jkowens:
This will be useful for disabling csrf protection in test suites or configuring the csrf key when the next version of rack-protection is released.

ie:

```
OmniAuth.config.request_validation_phase = OmniAuth::AuthenticityTokenProtection.new(allow_if: ->(env) { true })
```

or

```
OmniAuth.config.request_validation_phase = OmniAuth::AuthenticityTokenProtection.new(key: :_csrf_token)
```

### DIFF
--- a/lib/omniauth/authenticity_token_protection.rb
+++ b/lib/omniauth/authenticity_token_protection.rb
@@ -18,6 +18,8 @@ module OmniAuth
       react env
     end
 
+    alias_method :call, :call!
+
   private
 
     def deny(_env)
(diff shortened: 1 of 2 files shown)


BobbyMcWho:
Thanks @jkowens, I had been turning it off in test suites by setting `OmniAuth.config.request_validation_phase = nil` or `OmniAuth.config.request_validation_phase = proc {}`, but this makes sense to have as well. Going to release it in a patch version release since it honestly should have been present in the 2.0.0 release if I hadn't overlooked it.

BobbyMcWho (approved):

jkowens:
Cool, I didn't think of those options! Thanks for the quick release 💯

Issue #1032, opened 2021-02-02
New `#mock_call` implementation improperly runs authenticity token verification on all requests


jsdalton:
A recent change altered the behavior of omniauth when run in test mode. See below for details.

The change appears to have been introduced here https://github.com/omniauth/omniauth/commit/1b784ffa5f128bea1a22d7d26477f73bb6b3cd08 .

### Configuration

- Provider Gem: `omniauth-*`
- Ruby Version: `2.7.2`
- Framework: `Rails`
- Platform: `Mac`

### Expected Behavior

When calling an Omniauth strategy while in test mode, the mock checks and tests should be executed in a way that mirrors normal mode.

In regular mode it would look like this:

1. Execute `request_call` phase if `on_request_path?` is true: https://github.com/omniauth/omniauth/blob/v2.0.1/lib/omniauth/strategy.rb#L192
2. Within `request_call`, execute authenticity token verification: https://github.com/omniauth/omniauth/blob/v2.0.1/lib/omniauth/strategy.rb#L236

In other words, token verification only occurs _if_ we are on the request path.

Again the expectation is that test mode would behave the same way: only executing verification if we are on the request path.

### Actual Behavior

When test mode is true, `mock_call` is used early in the execution of the middleware. Unfortunately, the authenticity token verification happens differently than it does in normal mode.

See https://github.com/omniauth/omniauth/blob/v2.0.1/lib/omniauth/strategy.rb#L302 .

The verification happens _before_ the check to `on_request_path?` is made. As a result, the verification happens on very request, not just those that are on the request path provided by omniauth.

The problem is twofold:

1. Any request executed while in test mode that is not part of Omniauth and that does not require a CSRF token will fail (an example would be API calls that are part of the same application).
2. The behavior is different than normal/production mode, meaning developers might inadvertently rely on this addtional verification check which would _not_ be executed in outside of test mode.

### Steps to Reproduce

1. Set  `OmniAuth.config.test_mode = true`
2. Execute any non-GET request to any endpoint on the same application that is not on the Omniauth request path.
3. The `on_failure` block is triggered.

### Fix

For what it's worth the fix seems relatively straightforward: Move the request_validation_phase check...

* from here: https://github.com/omniauth/omniauth/blob/v2.0.1/lib/omniauth/strategy.rb#L302
* to here: https://github.com/omniauth/omniauth/blob/v2.0.1/lib/omniauth/strategy.rb#L315
[...]

Issue #1030, opened 2021-01-26
Rescue StandardError makes all development errors failures


pgwillia:
Since bumping omniauth from 1.9.1 to 2.0.1 in [our application](https://github.com/ualbertalib/jupiter) all development errors have become failures.  For example I purposefully introduce an error to a view.  It is caught by omniauth and calls fail which means that we can no longer use [better_errors](https://rubygems.org/gems/better_errors) to debug.  

```
195:       rescue StandardError => e
    196:        binding.pry
 => 197:         return fail!(e.message, e)
    198:       end
    199: 
    200:       @app.call(env)
    201:     end

[1] pry(#<OmniAuth::Strategies::SAML>)> e
=> #<ActionView::Template::Error: undefined local variable or method `oops' for #<#<Class:0x00007f533c178068>:0x000055e0bcf640a8>>
```

Is there something that we misconfigured in omniauth setup?  We catch the failures to give users a message in their browser
https://github.com/ualbertalib/jupiter/blob/fe8409b9d4b4345e409c91a0cf6744b9ce4b91ba/config/initializers/omniauth.rb#L22
We can make this conditional based on environment but wondering if there is a better way?

https://github.com/omniauth/omniauth/blob/40e354c71094b663f928e1a987977690b67f2aa4/lib/omniauth/strategy.rb#L190-L197

BobbyMcWho:
You’ll want to take a look at the `OmniAuth.config.on_failure` setting, by default it’s set to call the OmniAuth Failure Endpoint https://github.com/omniauth/omniauth/blob/master/lib/omniauth/failure_endpoint.rb

BobbyMcWho:
If you're looking to preserve raising the errors out in development I'd suggest something like 
```ruby 
on_failure do |env|
  if Rails.env.development?
    OmniAuth::FailureEndpoint.call(env)
  else
    SessionsController.action(:failure).call(env)
  end
end
```

pgwillia:
Yes.  That's effectively what we've done.  We still think it's strange that omniauth is intercepting all errors (not just the ones relevant to omniauth).

BobbyMcWho:
I'll look into it a bit more, it really should only be catching errors that occur during the auth process

trysmr:
Hi, I have a same problem. I use a omniauth-saml gem.

[...]

PR #1033, opened 2021-02-02
Ensure mock call verifies authenticity tokens with same logic as real call

jsdalton:
Readjusts when the authenticity token verification happens in `#mock_call` to align with the execution of the real call.

## Purpose

The current implementation of `mock_call` was verifying the token for all requests, regardless of whether the current path is on the omniauth request path. The change was introduced recently in https://github.com/omniauth/omniauth/commit/1b784ffa5f128bea1a22d7d26477f73bb6b3cd08. See https://github.com/omniauth/omniauth/issues/1032 for details.

This creates two problems:

1. When test mode is on, the authenticity verification logic is run inappropriately against requests where this may not even be wanted.
2. The behavior varies from actual production behavior, potentially allowing bugs to be introduced by unwary developers.

## Approach

* Move the request validation logic into `#mock_request_call `so that it tracks more closely with the real validation logic that is triggered from `#request_call`.
* Add a spec test to capture the regression

Please feel free to introduce this change in a different manner if needed. My main interest is unblocking a project I have where the newly introduced logic change is breaking many tests. We are disabling the `request_validation_phase` in all tests, but would like to re-enable it for integration tests to ensure our tests specify the behavior correctly.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -299,7 +299,6 @@ module OmniAuth
     # in test mode.
     def mock_call!(*)
       begin
-        OmniAuth.config.request_validation_phase.call(env) if OmniAuth.config.request_validation_phase
         return mock_request_call if on_request_path? && OmniAuth.config.allowed_request_methods.include?(request.request_method.downcase.to_sym)
         return mock_callback_call if on_callback_path?
       rescue StandardError => e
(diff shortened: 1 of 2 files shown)


BobbyMcWho (approved):
Appreciate the issue report and PR @jsdalton, I think this also may relate to #1030

BobbyMcWho:
This has been tagged and released to rubygems as v2.0.2

jsdalton:
Thanks for the quick merge and update @BobbyMcWho. 🙇

PR #1035, opened 2021-02-12
Better handle errors that come from the actual app.

### LINKED ISSUE #1030: Rescue StandardError makes all development errors failures

BobbyMcWho:
As a consequence of the changes that were merged in #689, errors
thrown by strategies that utilize other_phase (or more specifically
call_app!), would be caught by omniauth, causing headaches for folks
looking to have those errors handled by their application. This
should allow for errors that come from the app to pass through, while
passing errors that come from the authentication phases to the fail!
handler.

Resolves #1030

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -193,6 +193,8 @@ module OmniAuth
         return callback_call if on_callback_path?
         return other_phase if respond_to?(:other_phase)
       rescue StandardError => e
+        raise e if env.delete('omniauth.error.app')
+
         return fail!(e.message, e)
       end
 
@@ -302,6 +304,8 @@ module OmniAuth
         return mock_request_call if on_request_path? && OmniAuth.config.allowed_request_methods.include?(request.request_method.downcase.to_sym)
         return mock_callback_call if on_callback_path?
       rescue StandardError => e
+        raise e if env.delete('omniauth.error.app')
+
         return fail!(e.message, e)
       end
 
@@ -462,6 +466,9 @@ module OmniAuth
 
     def call_app!(env = @env)
       @app.call(env)
+    rescue StandardError => e
+      env['omniauth.error.app'] = true
+      raise e
     end
 
     def full_host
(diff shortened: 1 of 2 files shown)


stnert (approved):

lib/omniauth/strategy.rb
```diff
@@ -462,6 +466,9 @@ def query_string
 
     def call_app!(env = @env)
       @app.call(env)
+    rescue StandardError => e
```
pgwillia, line 469: I was wondering if it would be possible to rescue a subclass of `StandardError` like `OmniAuthError` or something like that.  I guess the challenge is that `call_app` is called from the other dependent gems (omniauth-saml, etc) and you can't really know what will be raised.

pgwillia (approved):
This works for me.  I can't think of a situation where the environment variable solution shouldn't work.  The tests illustrate the exact situation that we experienced. Thanks @BobbyMcWho

PR #1041, opened 2021-03-31
Warn only on GET requests for a login path

charlie-wasp:
This is a fix for warning logging brought up in https://github.com/omniauth/omniauth/pull/1010#discussion_r604382856

The problem is that currently warning message is logged on *every* GET request, not only for a login path. Such messages can bloat application logs very fast

Also, it contradicts this particular piece of documentation ([source](https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0#get)):

> Because using GET for login poses concerns (see OWASP Cheatsheet), after upgrading OmniAuth **will log a :warn level log with every GET request to a login path** when your OmniAuth.config.allowed_request_methods include :get.

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -180,9 +180,10 @@ module OmniAuth
         raise(error)
       end
 
-      warn_if_using_get
-
       @env = env
+
+      warn_if_using_get_on_request_path
+
       @env['omniauth.strategy'] = self if on_auth_path?
 
       return mock_call!(env) if OmniAuth.config.test_mode
@@ -201,7 +202,8 @@ module OmniAuth
       @app.call(env)
     end
 
-    def warn_if_using_get
+    def warn_if_using_get_on_request_path
+      return unless on_request_path?
       return unless OmniAuth.config.allowed_request_methods.include?(:get)
       return if OmniAuth.config.silence_get_warning
 
--- a/spec/omniauth/strategy_spec.rb
+++ b/spec/omniauth/strategy_spec.rb
@@ -986,6 +986,9 @@ describe OmniAuth::Strategy do
       end
 
       context 'with allowed GET' do
+        let(:path) { '/auth/test' }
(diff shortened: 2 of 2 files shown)


charlie-wasp:
There is a couple of failed jruby builds, and I actually don't know, what can I do about them. They fail yet on the ruby setup, my code isn't involved, I could guess

sponomarev (approved):
👍

BobbyMcWho (changes requested):
Sorry for the delay on this, was there something that you found during testing that made `on_request_path?` a better choice than `on_auth_path?`. I'd lean towards using the latter if not

charlie-wasp:
@BobbyMcWho no particular research was made on this matter. Reason for choosing `request_path` is only [current docs](https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0#get) – it says *"login path"*, and `on_auth_path?` returns `true` for callback path too AFAIK

BobbyMcWho:
> @BobbyMcWho no particular research was made on this matter. Reason for choosing `request_path` is only [current docs](https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0#get) – it says _"login path"_, and `on_auth_path?` returns `true` for callback path too AFAIK

I think we will want to log on the callback path as well, so I'd prefer the `on_auth_path?`

charlie-wasp:
@BobbyMcWho but original [CVE](https://nvd.nist.gov/vuln/detail/CVE-2015-9284) states, that **request path** is vulnerable to CSRF attacks, it says nothing about callbacks. Besides that, correct me, if I'm wrong, but GET-request is acceptable for callback URI, since user is redirected there?

BobbyMcWho (approved):

BobbyMcWho:
This has been released in [v2.0.4](https://github.com/omniauth/omniauth/releases/tag/v2.0.4)

charlie-wasp:
Thanks @BobbyMcWho! Really appreciated such a quick feedback

PR #1056, opened 2022-02-08
Update README.md

bradgessler:
I spent an hour trying to figure out why `/auth/developer` wasn't working from the README, and its because the `:get` method wasn't allowed for `OmniAuth.config.allowed_request_methods`

Is there a better way to do this per provider instead of globally for all providers? That would be ideal, but if not, what's the danger of adding `:get` to the default list vs the change I'm proposing?

Regardless of the path, this change will prevent future devs from spending lots of time trying to figure out why `/auth/developer` doesn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,10 @@ example that you might put into a Rails initializer at
 
 ```ruby
 Rails.application.config.middleware.use OmniAuth::Builder do
-  provider :developer unless Rails.env.production?
+  unless Rails.env.production?
+    provider :developer
+    OmniAuth.config.allowed_request_methods = [:get, :post]
+  end
   provider :twitter, ENV['TWITTER_KEY'], ENV['TWITTER_SECRET']
 end
 ```


BobbyMcWho:
Hi @bradgessler, there's a lot of history w/ using GET, and a long standing vulnerability that was fixed last year around it. https://github.com/advisories/GHSA-ww4x-rwq6-qpgf

I don't particularly wish to recommend folks to use GET, even in dev, since it's not recommended to use in prod. Routes should be set up using a post, and there are examples in the wiki on how to set it up. 

If the readme isn't clear as far as setting up the post button, we should clarify that

bradgessler:
Makes sense, I figured there was a security issue. If that's the case I'd simply eliminate `:get`

Based on your response, the issue is that the `developer` strategy won't work unless `:get` is allowed, which I configured in the `README` for the development environment. At this point, I won't be able to meaningfully contribute to this PR since this seems to be a question for the maintainers of this project.

The issue stands that as documented (prior to this PR), the `/auth/developer` strategy doesn't work out of the box and fails silently. I suppose one option would be to show an error message on a `:get` request w/ the proper HTTP code.

Issue #1089, opened 2022-09-26
Compatibility issues with rack-protection 3.x


deepfryed:
Please complete all sections.

### Configuration

- Provider Gem: `omniauth`
- Ruby Version: `3.1.2`
- Framework: `N/A`
- Platform: `Linux`

### Expected Behavior

The library should load

### Actual Behavior

load error, 'rack-protection' not found

### Steps to Reproduce

Setup a ruby or rails app with rack-protection 3.x. The file 'rack-protection.rb' has been renamed to 'rack_protection.rb' in sinatra.

BobbyMcWho:
Hi, it appears that Sinatra has merged a fix to keep backwards compatibility

BobbyMcWho:
If they haven't cut a release, you could point your gem version at their master sha

BobbyMcWho:
Looks like 3.0.1 was pushed 3 minutes ago

biologistbrian:
Yeehaw! just saw that. yay!

PR #1090, opened 2022-09-26
Support rack-protection 3.x

hieuk09:
Fix https://github.com/omniauth/omniauth/issues/1089

### DIFF
--- a/lib/omniauth/authenticity_token_protection.rb
+++ b/lib/omniauth/authenticity_token_protection.rb
@@ -1,4 +1,4 @@
-require 'rack-protection'
+require 'rack/protection'
 
 module OmniAuth
   class AuthenticityError < StandardError; end


mvz:
I don't think the maintainers of rack-protection intended this breaking change, so it might be useful to instead request them to add back `rack-protection.rb`.

hieuk09:
Thanks for taking a look at this. I'll wait for answer from Sinatra folks then. I'll close this PR if `rack-protection` reverts the breaking change.

lib/omniauth/authenticity_token_protection.rb
```diff
@@ -1,4 +1,9 @@
-require 'rack-protection'
+begin
+  require 'rack-protection'
+rescue LoadError
+  # support rack/protection 3.x
+  require 'rack_protection'
+end
```
dentarg, line 6: Have you tried just a single require `require 'rack/protection`? Might work for both versions

dentarg: And wouldn't requiring `rack_protection` try to load the wrong gem? (If https://rubygems.org/gems/rack_protection would exist (hopefully it doesn't exist because the name is protected from squatting))

hieuk09: Yes, that should work. Thanks for the suggestion.

mvz: > And wouldn't requiring `rack_protection` try to load the wrong gem? (If https://rubygems.org/gems/rack_protection would exist (hopefully it doesn't exist because the name is protected from squatting))

It's worse than that: any gem could ship a file `rack_protection.rb` or `rack-protection.rb`.

hieuk09:
Close this PR as `rack-protection` gem reverts their change.

PR #1093, opened 2022-10-04
Rename Builder to AuthBuilder and create Builder wrapper

### LINKED ISSUE #1083: `Rack::Builder#to_app` is not designed to be called for every request.
https://github.com/omniauth/omniauth/blob/9cd7aad3a0bd0f27a6de81973a17722e09d4a7d2/lib/omniauth/builder.rb#L44

It can have less than ideal performance implications.

speckins:
Previously, OmniAuth::Builder was a subclass of Rack::Builder. Rack::Builder#call calls #to_app, rebuilding the Rack application on each request, and, as it says in the Rack::Builder comments, this may have a negative impact on performance.

This commit renames the original OmniAuth::Builder to OmniAuth::AuthBuilder and replaces it with a wrapper to be used as middleware (as is suggested in the documentation).

Using OmniAuth::Builder will now only build the application once.

Resolves #1083.


speckins:
I didn't undef the `#call` method from `AuthBuilder` completely as I suggested in the issue; let me know if you want to go that far.

Also, do you think we should add some kind of "signpost" in the `Builder` code to indicate to downstream users that the class has been renamed?  Maybe adding versions of `#provider`, `#options`, etc., that raise with a message, or maybe just a comment in the code?

PR #1095, opened 2022-10-13
Add bake-test-external for devise

### LINKED ISSUE #1091: Add Bake test for devise
https://github.com/omniauth/omniauth/issues/1083#issuecomment-1266106100

QWYNG:
fix #1091

Add bake-test-external and test devise.

see also

https://github.com/lostisland/faraday/pull/1439


ioquatix:
This is a good idea.

PR #1106, opened 2023-03-03
   Use GET with :developer strategy.

### LINKED ISSUE #957: developer strategy: ActionController::InvalidAuthenticityToken in SessionsController#create
Please complete all sections.

### Configuration

- Provider Gem: `omniauth*`     #.lock shows omniauth (1.9.0)
- Ruby Version: `2.4.0`
- Framework: 'rails', '4.2.10'
- Platform: `ubuntu`   #on cloud 9 ide

### Expected Behavior
Using develper strategy, callback should be provided with valid token

### Actual Behavior
Goes to auth/developer login  form correctly, but then raises following error instead of returning to auth/developer/callback which routes to sessions#create:
  ActionController::InvalidAuthenticityToken in SessionsController#create
  ActionController::InvalidAuthenticityToken

### Steps to Reproduce
Register or Log In with Developer
Click on above link which is on login page.

Register or Log In with GitHub
Register or Log In with Facebook
Other links work normally, also passes Cucumber tests in test mode.
Only problem is that I need to do an end run on actual authentication while I'm in develper mode so actual providers aren't abused.

Let me know which files I should include.

### LINKED ISSUE #1061: Developer provider not working under Rails 7.0.2.3
Please complete all sections.

### Configuration

- Provider Gem: `omniauth-2.0.4`
- Ruby Version: `2.7.5`
- Framework: `rails 7.0.2.3`
- Platform: `ubuntu`

### Expected Behavior

Gemfile
```
gem 'omniauth'
gem "omniauth-rails_csrf_protection"
```

config/initializers/omniauth.rb
```
  provider :developer
end
```

app/views/sessions/new.html.erb 
```
<%= form_tag('/auth/developer', method: 'post') do %>
  <button type='submit'>Login with Developer</button>
<% end %>
```

Clicking on `Login` should render the developer form and log you in using fake credentials.

### Actual Behavior

The only thing in the log is:
```
Started POST "/auth/developer" for 127.0.0.1 at 2022-03-21 21:58:57 +0100
D, [2022-03-21T21:58:57.950088 #3941647] DEBUG -- : (developer) Request phase initiated.
```

So the middleware is hit, but there is nothing after this.

### Steps to Reproduce

`gem install rails --version 7.0.2.3`
`rails new --css bootstrap`
Add the code in expected behaviour.
`bundle install`
`rails s`
`click Login`

### LINKED ISSUE #1087: CSRF in Rack for callbacks requests
Please complete all sections.

### Configuration

- Provider Gem: `omniauth-developer`
- Ruby Version: `3.1.2`
- Framework: [`Flame`](https://github.com/AlexWayfer/flame)
- Platform: `Arch Linux`

### Expected Behavior

OmniAuth-Developer form contains CSRF param and passes it to `/auth/developer/callback`.

### Actual Behavior

I see no `_csrf` param, even when completed setup instructions for compatibility.

### Steps to Reproduce

```ruby
::OmniAuth::AuthenticityTokenProtection.default_options(key: 'csrf.token', authenticity_param: '_csrf')

use Rack::Csrf, raise: true

use OmniAuth::Builder do
	provider :developer
end
```

### Additional info

I've tried `allow_if` block from #1074, but it seems never called.

The form to authenticate:

```
<form action="/auth/developer" method="POST">
	<input type="hidden" name="_csrf" value="Bia4XjpqNYzblbtv2TtiYqnsniZIQIHoMZJo5b4eqJI">
	<button type="submit">Connect Developer</button>
</form>
```

The form from the OmniAuth-Developer provider:

image

Without `_csrf` param (hidden?) the `/callback` will be rejected (by `Rack::Csrf` middleware).

cycomachead:
This changes the default method of the :developer strategy to GET. It does
   this by allowing OmniAuth::Form to accept a `:method` option, which is
   optional and defaults to 'post', the current behavior. Because the default
   behavior remains unchanged, I don't expect this to introduct any breaking
   changes.

   This allows the developer strategy to work with the directions provided in
   the README. Further, it seems that the default action of most stratigies is
   to use GET for their callbacks.

   Fixes #1087 Fixes #1061 Fixes #957

### DIFF
--- a/lib/omniauth/form.rb
+++ b/lib/omniauth/form.rb
@@ -7,6 +7,7 @@ module OmniAuth
     def initialize(options = {})
       options[:title] ||= 'Authentication Info Required'
       options[:header_info] ||= ''
+      options[:method] ||= 'post'
       self.options = options
 
       @html = +'' # unary + string allows it to be mutable if strings are frozen
@@ -75,7 +76,7 @@ module OmniAuth
       </head>
       <body>
       <h1>#{title}</h1>
-      <form method='post' #{"action='#{options[:url]}' " if options[:url]}noValidate='noValidate'>
+      <form method='#{options[:method]}' #{"action='#{options[:url]}' " if options[:url]}noValidate='noValidate'>
       HTML
       self
     end
--- a/lib/omniauth/strategies/developer.rb
+++ b/lib/omniauth/strategies/developer.rb
@@ -35,7 +35,7 @@ module OmniAuth
       option :uid_field, :email
 
       def request_phase
-        form = OmniAuth::Form.new(:title => 'User Info', :url => callback_path)
+        form = OmniAuth::Form.new(:title => 'User Info', :url => callback_path, :method => 'get')
         options.fields.each do |field|
           form.text_field field.to_s.capitalize.tr('_', ' '), field.to_s
         end
--- a/spec/omniauth/form_spec.rb
+++ b/spec/omniauth/form_spec.rb
@@ -20,6 +20,14 @@ describe OmniAuth::Form do
     it 'sets an H1 tag from the passed :title option' do
       expect(OmniAuth::Form.new(:title => 'Something Cool').to_html).to be_include('<h1>Something Cool</h1>')
     end
+
+    it 'sets the default form method to post' do
+      expect(OmniAuth::Form.new.to_html).to be_include("method='post'")
+    end
+
+    it 'sets the form method to the passed :method option' do
+      expect(OmniAuth::Form.new(:method => 'get').to_html).to be_include("method='get'")
+    end
   end
 
   describe '#password_field' do
--- a/spec/omniauth/strategies/developer_spec.rb
+++ b/spec/omniauth/strategies/developer_spec.rb
@@ -31,7 +31,7 @@ describe OmniAuth::Strategies::Developer do
 
     context 'with default options' do
       before do
-        post '/auth/developer/callback', :name => 'Example User', :email => 'user@example.com'
+        get '/auth/developer/callback', :name => 'Example User', :email => 'user@example.com'
       end
 
       it 'sets the name in the auth hash' do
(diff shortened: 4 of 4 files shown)


BobbyMcWho:
I'll get around to diagnosing the POST issue, but GET is not recommended for omniauth strategies as it is vulnerable to CVE-2015-9284

cycomachead:
I am a bit confused though. The CVE discusses a POST to launch to the authentication provider, not the callback route. Today, as best I understand - and it is still true with this PR:* user clicks a POST form to be redirected to a login provider* provider does its stuff* provider redirects to /auth/:provider/callback via a GET* Omni auth/each state get validates the paeans to that request. Today, in Rails, if you want to accept the POST in the callback, you need to add a 2nd route not mentioned in the docs, and you need to add an controller hook to skip the token verification - which I think is understandable since it’s very easy for that hook to be too broad (though it likely won’t be a security issue since we do allow GETs). But unless I am wrong, basically every strategy I’ve used is fine with a GET callback - google, GitHub, Microsoft, a bunch of others. There are two other ways I could see fixing this:* the developer strategy from POSTs to some new internal route which is NOT the callback route. That route then redirects to the callback route with a GET. * we allow the rails CSRF gem to inject a CSRF token, generated by Rails unless the appropriate Rails config, so that Rails can validate it. You still have to configure your app to accept a POST callback, which feels odd just for development but it would work. Thanks for considering this! -- Michael BallFrom my iPhonemichaelball.coOn Mar 3, 2023, at 5:36 AM, Bobby McDonald ***@***.***> wrote:﻿
Closed #1106.

BobbyMcWho:
Sorry, you're correct, I only glanced at this PR and assumed it was altering the POST allowed actions

cycomachead:
No worries, thanks! I mean, the Development provider is definitely insecure as is, but hopefully it is clear this shouldn’t be used in production! Still, if there’s a better path, I’m happy to help. Sent from my iPadOn Mar 3, 2023, at 10:00 AM, Bobby McDonald ***@***.***> wrote:﻿
Sorry, you're correct, I only glanced at this PR and assumed it was altering the POST allowed actions

BobbyMcWho (approved):

spryffee:
When there will be next stable release that includes this change?

salzig:
I would appreciate a 2.1.2 which contains this change.

BobbyMcWho:
> I would appreciate a 2.1.2 which contains this change. 

Released

cycomachead:
Thanks very much! :)

PR #1105, opened 2023-02-17
Fix specs on CI against TruffleRuby 22.3.1

andrykonchin:
Changes:
- use the latest `bundler` version

It should be merged after https://github.com/omniauth/omniauth/pull/1104 to not force using an old `bundler` version

### Notes

Looks like there was an issue in `bundler` version vendored with TruffleRuby 22.3.1.

It leads to spec failures on omniauth CI ([build](https://github.com/omniauth/omniauth/actions/runs/4205385104/jobs/7298340176)):

```
An error occurred while loading ./spec/omniauth/auth_hash_spec.rb. - Did you mean?
                    rspec ./spec/omniauth/strategy_spec.rb
                    rspec ./spec/omniauth/builder_spec.rb

Failure/Error: require 'hashie/mash'

LoadError:
  cannot load such file -- hashie/mash
# ./lib/omniauth/key_store.rb:1:in `<top (required)>'
# ./lib/omniauth/auth_hash.rb:1:in `<top (required)>'
# ./lib/omniauth.rb:4[7](https://github.com/omniauth/omniauth/actions/runs/4205385104/jobs/7298340176#step:5:8):in `defaults'
# ./lib/omniauth.rb:53:in `initialize'
# /home/runner/.rubies/truffleruby-22.3.1/lib/mri/singleton.rb:[12](https://github.com/omniauth/omniauth/actions/runs/4205385104/jobs/7298340176#step:5:13)7:in `block in instance'
# /home/runner/.rubies/truffleruby-22.3.1/lib/mri/singleton.rb:125:in `synchronize'
# /home/runner/.rubies/truffleruby-22.3.1/lib/mri/singleton.rb:125:in `instance'
# ./lib/omniauth.rb:[13](https://github.com/omniauth/omniauth/actions/runs/4205385104/jobs/7298340176#step:5:14)1:in `config'
# ./spec/helper.rb:27:in `<top (required)>'
# ./spec/omniauth/auth_hash_spec.rb:1:in `<top (required)>'
```

The issue looks like was fixed in `bundler` and running specs with current version works fine.

PR #1104, opened 2023-02-17
Fix `bundle install` step on CI for TruffleRuby

andrykonchin:
Currently CI builds agains TruffleRuby 22.3.1 are failed on the `bundle install` step.
```
...
Fetching webrick 1.7.0
  Installing webrick 1.7.0
  Fetching yard 0.9.28
  Installing yard 0.9.28
  --- ERROR REPORT TEMPLATE -------------------------------------------------------
  
...
  NoMethodError: undefined method `full_name' for nil:NilClass
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:124:in `block (2 levels) in check_for_unmet_dependencies'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:123:in `each'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:123:in `block in check_for_unmet_dependencies'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:122:in `each'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:122:in `check_for_unmet_dependencies'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:[100](https://github.com/omniauth/omniauth/actions/runs/3970157787/jobs/6805553917#step:3:106):in `call'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer/parallel_installer.rb:71:in `call'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer.rb:262:in `install_in_parallel'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer.rb:209:in `install'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/installer.rb:89:in `block in run'
    /home/runner/.rubies/truffleruby-22.3.0/lib/mri/bundler/process_lock.rb:12:in `block in lock'
    <internal:core> core/io.rb:670:in `open'
```
https://github.com/omniauth/omniauth/actions/runs/3970157787/jobs/6805553917

 I assume that the issue is related to invalid/incorrect cached `Gemfile.lock` file as far as locally I cannot reproduce the issue.

[...]

PR #1109, opened 2023-03-11
ci: use ubuntu-latest runner

nschonni:
The 18.04 runners are depriciated https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/
I noticed some of the jobs had OS matrixes, but they weren't wired in, so I updated them to actually use those runner

### DIFF
--- a/.github/workflows/truffle_ruby.yml
+++ b/.github/workflows/truffle_ruby.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/truffle_ruby.yml
+++ b/.github/workflows/truffle_ruby.yml
@@ -19,7 +19,6 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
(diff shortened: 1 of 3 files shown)


eregon:
:+1: It would be nice to merge because many jobs don't run now, e.g. https://github.com/omniauth/omniauth/actions/runs/5068837971/jobs/9101646696

.github/workflows/truffle_ruby.yml
```diff
@@ -19,7 +19,6 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
-    - name: Install dependencies
-      run: bundle install
+        bundler-cache: true
```
andrykonchin, line 22: This line reverts a fix in https://github.com/omniauth/omniauth/pull/1104

eregon: It seems to pass now: https://github.com/omniauth/omniauth/actions/runs/5129258050/jobs/9226698922
So maybe no longer necessary?

andrykonchin: Right, I also have closed another fix for 22.3 - https://github.com/omniauth/omniauth/pull/1105.

PR #1117, opened 2023-05-30
ci: Merge JRuby/Truffleruby jobs to main CI

nschonni:
Since the jobs were essentially the same (minis one ENV value), thought they could be combined. Because of the change of "name", need to varify it doesn't break the required jobs configured for branch protection


BobbyMcWho:
I had separated them I believe so that there could be separate README badges and the frequently failing jruby and truffleruby didn't make the main build failing

nschonni:
Fair enough

PR #1119, opened 2023-05-30
ci: Add Rubocop checks to CI and use latest versions

nschonni:
Noticed that it wasn't currently called in the CI, so I updated to the latest version, and then added a job in the CI to make sure it's called during the PR process


BobbyMcWho:
I honestly don't strictly enforce rubocop on this repo any more, it's not under super active development, so it's a bit more hassle than it's worth

nschonni:
@BobbyMcWho my thinking was for the automatic rewrites it can do if the minimum Ruby version get's bumped to v3 when updating `TargetRubyVersion`

PR #1120, opened 2023-05-30
chore: cleanup coveralls/simplecov setup

nschonni:
The coveralls_reborn mentions that the gem isn't needed for Actions.
Followed setup from https://github.com/coverallsapp/github-action/issues/29#issuecomment-704366557


nschonni:
This doesn't fix the JRuby failures, but it get's rid of some unrelated errors like https://github.com/omniauth/omniauth/actions/runs/5125273149/jobs/9218237165?pr=1119#step:4:210 when it tries to load some of that stuff at the end of the runs

PR #1134, opened 2024-07-19
Do not override `omniauth.origin` in environment in test mode

### LINKED ISSUE #934: mocking `request.env['omniauth.origin']` when request.
Hi I would like to rspec testing with `request.env['omniauth.origin']`. I successfully mocked `OmniAuth.config.mock_auth[:google_oauth2]`. but I can't set the `request.env['omniauth.origin']` value. The callback controller doesn't know it. How can i do it?

### Configuration

- Provider Gem: `omniauth-google-oauth`
- Ruby Version: `2.4.2`
- Framework: `rails`

### RSpec
```
it 'should redirect to another page' do
      valid_google_oauth2_login_setup
      get '/auth/google_oauth2/callback'
      request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:google_oauth2]
      request.env['omniauth.origin'] = 'http://www.xxxx.com'
end
```

TastyPi:
I'm attempting to write a test for our authentication callback Rails controller, however using `ActionDispatch::IntegrationTest` it is not possible to modify the `session`. This change makes it easy to test.

Fixes #934

Example with Rails:

```ruby
class ExampleController < ApplicationController
  def callback
    # Add user to session
    redirect_to(env["omniauth.origin"] || "/")
  end
end

class ExampleControllerTest < ActionDispatch::IntegrationTest
  test "redirects to origin" do
    get "/auth/example/callback", env: {"omniauth.origin" => "/foo"}

    assert_redirected_to "/foo"
  end
end
```

### DIFF
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -335,7 +335,9 @@ module OmniAuth
 
     def mock_callback_call
       setup_phase
-      @env['omniauth.origin'] ||= session.delete('omniauth.origin')
+
+      origin = session.delete('omniauth.origin')
+      @env['omniauth.origin'] ||= origin
       @env['omniauth.origin'] = nil if env['omniauth.origin'] == ''
       @env['omniauth.params'] = session.delete('omniauth.params') || {}
 
--- a/lib/omniauth/strategy.rb
+++ b/lib/omniauth/strategy.rb
@@ -335,7 +335,7 @@ module OmniAuth
 
     def mock_callback_call
       setup_phase
-      @env['omniauth.origin'] = session.delete('omniauth.origin')
+      @env['omniauth.origin'] ||= session.delete('omniauth.origin')
       @env['omniauth.origin'] = nil if env['omniauth.origin'] == ''
       @env['omniauth.params'] = session.delete('omniauth.params') || {}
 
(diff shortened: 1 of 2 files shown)


lib/omniauth/strategy.rb
```diff
@@ -335,7 +335,7 @@ def mock_request_call
 
     def mock_callback_call
       setup_phase
-      @env['omniauth.origin'] = session.delete('omniauth.origin')
+      @env['omniauth.origin'] ||= session.delete('omniauth.origin')
```
BobbyMcWho, line 338: This has the side effect of leaving the omniauth.origin in the session if it's already been set

TastyPi: Good point, updated

TastyPi:
@BobbyMcWho I've addressed your comment, could you take another look please?

BobbyMcWho (approved):

TastyPi:
@BobbyMcWho the test failures are unrelated to this change.

The `head` and `debug` failures are complaining that `URI::ABS_URI` does not exist, which is being addressed by https://github.com/omniauth/omniauth/pull/1136.

The `(macos, 2.5)` failure appears to be because CRuby does not support `macos-arm64`. `macos-latest` is now arm64 instead of x86 https://github.com/actions/runner-images?tab=readme-ov-file#available-images. I've created https://github.com/omniauth/omniauth/pull/1137 to address this failing pair.

TastyPi:
@BobbyMcWho could we get a new Omniauth release that includes this change? Looks like there hasn't been a new release in nearly a year.

